PR #77, opened 2026-08-28
Release QA protocol: prove every cache, gate the tag on it

janicduplessis:
field-test-protocol.md 118 -> 637 lines: zero-config proof per platform, every cache with three proofs (engaged/stores/reused), release builds, flavors, the four launched values, severity rules, and a release QA matrix saying which repo shapes a patch vs a minor requires. RELEASE.md gains a pre-tag QA gate. Also fixes docs that were wrong: pre-flight referenced packages/rn-iso/bin/cli.js (does not exist), 'all three packages' (four since core), and e2e-and-ci.md described npm/node:test/Node 20.


janicduplessis:
Closing as superseded after auditing this branch against current main. The core idea—a pre-tag QA gate with evidence-backed release claims—still makes sense and is tracked in #91, but this implementation cannot be safely rebased: it targets the retired rn-iso package and .rn-iso project-local state, describes obsolete Node/CJS/runtime behavior, assigns semantics to reserved launched=false, and manually duplicates cache checks now covered by the native caches suite. A focused rewrite from current main is safer than resolving the three large doc conflicts or mechanically renaming this 637-line protocol.

PR #141, opened 2026-08-31
Install on a connected physical Android device with --device

### LINKED ISSUE #140: Support installing on a connected physical Android device
`stim android` can only install on the owned emulator. There is no way to run a build on a phone plugged into the same machine, so any check that needs real hardware -- a camera, a physical keyboard, a device-specific layout bug, an OEM quirk -- has to leave Stim and drive Gradle and adb by hand.

The v3 surface removed the old `--serial` flag along with the rest of the physical-device path, and the invariants in `AGENTS.md` have forbidden it since. That was the right call for ownership: Stim must not create, boot, shut down, or delete somebody's phone. But the ownership rule and the install path are separable. Everything downstream of the build already takes a serial and does not care what kind of device it belongs to: `installAndroidApp`, `launchAndroidApp`, `reverseMetroPorts`, `verifyLaunch`, and the logcat collector.

Proposal: `stim android --device [serial]` builds as usual, then installs and launches on a connected physical device. With no serial it uses the one connected device, and refuses with the candidate list when adb reports several.

The physical path skips everything that manages an owned device -- the device-capacity check, AVD creation, the boot wait -- and writes no device record, so `stop`, `gc`, and `teardown.ts` cannot reach the hardware. The build, the fingerprint, the build cache, and the Metro port gate stay as they are.

One bug blocks this today. `engine/app-install.ts` hardcodes the emulator host loopback `10.0.2.2` for both `debug_http_host` and the dev-client deep link. That address means nothing on a phone, so the app would find no bundler. On hardware the host must be `localhost`, which the `adb reverse` that already runs serves.

Out of scope: physical iOS devices, which need code signing and provisioning profiles.

janicduplessis:
Closes #140

## Description

`stim android` could only install on the owned emulator, so anything needing real hardware meant leaving Stim and driving Gradle and adb by hand.

Physical-device support was removed deliberately, in [32b96b2](https://github.com/appandflow/stim/commit/32b96b2daaf14a80311932d58bd706b10b09852e), which deleted the old `--serial` flag and wrote the ban into invariants 2 and 3 of `AGENTS.md`. That was right about ownership -- Stim must not create, boot, shut down, or delete somebody's phone -- but it conflated owning a device with installing on one. This PR keeps the ownership half and drops the rest, so those two invariants are rewritten rather than ignored.

## Solution

`stim android --device [serial]` builds as usual, then installs and launches on a connected physical device. With no serial it uses the one connected device, and refuses with the candidate list when adb reports several. It cannot be combined with `--remote`.

**`debug_http_host` and the dev-client deep link now resolve to `localhost` on a physical device instead of `10.0.2.2`.** This is a fix, not a preference: the emulator host loopback is not routable from a phone, so without it the app finds no bundler and loads no JS at all. The `adb reverse` that already ran serves `localhost`. Emulators keep `10.0.2.2`.

Stim never records the phone as a device, so `stop`, `gc`, and `teardown.ts` -- all of which act on the project's device record -- have nothing to act on. The physical branch also skips the device-capacity check, AVD creation, and the boot wait, but keeps the build-slot cap, since the build runs locally either way.

A device that reports itself as an emulator is refused. Genymotion and
`adb connect` emulators appear to adb as `host:port` rather than `emulator-N`, so
serial shape alone would have let `--device` install on a user-created emulator.
`ro.kernel.qemu` and `ro.hardware` settle it.

`--variant` is unchanged; a flavored project still needs it, physical or not.

## One device, many workspaces

Every workspace gets its own emulator, but they all share the one phone, so it is
worth being precise about what does and does not collide.

Two `adb reverse` pairs are registered: `<port> -> <port>`, which is what an app
with `debug_http_host` written actually uses, and `8081 -> <port>`, the fallback
for an app whose prefs write failed. Only the first is per-workspace. Two
workspaces on 8082 and 8083 therefore do not interfere: each app follows its own
`debug_http_host` to its own Metro. Registering a device port replaces any
previous mapping for it, so the shared `8081` entry does belong to whichever
workspace ran last -- but that only decides where an app lands when its prefs
write failed, which is already a warned, degraded state.

**Known limitation:** two workspaces building the same `applicationId` overwrite
each other's install, because that is one package name and one device. The last
`--device` run wins. Emulator isolation does not extend to hardware, and this PR
does not pretend otherwise.

## Risk

Contained. Nothing changes for an emulator run: the physical work is behind `--device`, and the host selection defaults to the existing `10.0.2.2`. The one shared edit is the new `physical` parameter threaded through `writeDebugHttpHost`, `androidDevClientUrl`, and `launchAndroidApp`, all defaulting to `false`.

The exposure worth naming is that Stim now issues `adb install`, `am start`, and `run-as` against hardware it does not own. Each is scoped to a serial adb already reports as `device`, and the destructive-uninstall retry on a signature conflict is still gated on release variants only, as before. No new adb verb reaches a physical device that did not already reach an emulator.

## Test plan

Verified end to end against a physical Samsung SM-G996W (`RFCR7081Q9L`), building tlon-apps `develop`:

```
device      SM-G996W (RFCR7081Q9L) connected, not owned by Stim
install     app-preview-debug.apk (37.9s)
launch      io.tlon.groups.preview (expo-dev-client deep link)
wired       debug_http_host localhost:8082 + adb reverse tcp:8081 -> tcp:8082
verify      ready: bundle loaded, stable for 3s, process alive (3s total)
```

`verify ready: bundle loaded` is the load-bearing line -- it is Metro actually serving the phone over the reverse, which is what the `10.0.2.2` to `localhost` change buys.

After the run, the project's entry in `~/.stim/config.json` has no `platforms` record at all and `stim status` prints no android line, so `stop` and `gc` have nothing to act on.

Also exercised the real `adb -s <serial> shell getprop ro.product.model` call behind `physicalDeviceModel`, per invariant 9, and confirmed real `adb devices` output parses to the fixtures the unit tests use.

### DIFF
--- a/packages/stim-cli/src/commands/android.ts
+++ b/packages/stim-cli/src/commands/android.ts
@@ -84,6 +84,7 @@ import {
   unverifiedLaunchLines,
   verifyAndroidReleaseLaunch,
   androidAppProcess,
+  installConflictKind,
   verifyLaunch,
 } from '../engine/app-install.ts';
 import {
@@ -94,6 +95,7 @@ import {
   findBuildTool,
   listAdbDevices,
   physicalDeviceModel,
+  probeEmulatorSerial,
   resolvePhysicalDevice,
 } from '../sim/android.ts';
 import { checkDeviceCapacity, ensureBooted, ensureOwnedDevice } from '../engine/device.ts';
@@ -632,6 +634,7 @@ interface RunAndroidOptions {
   device?: string | boolean | null;
   listDevices?: typeof listAdbDevices;
   deviceModel?: typeof physicalDeviceModel;
+  isEmulatorDevice?: typeof probeEmulatorSerial;
   readApkPackage?: (apkPath: string | null) => string | null;
   remoteDevice?: RemoteDeviceBackend | null;
   resolveSettingsFor?: typeof resolveSettings;
@@ -716,6 +719,7 @@ interface VerifyAndroidRunArgs {
   launchedAt: number;
   metroPort: number | null;
   isExpo: boolean;
+  physical: boolean;
   scheme?: string | null;
   phase: (label: unknown, text: string) => void;
 }
@@ -733,6 +737,7 @@ async function verifyAndroidRun({
   launchedAt,
   metroPort,
   isExpo,
+  physical,
   scheme,
   phase,
 }: VerifyAndroidRunArgs): Promise<boolean | string> {
@@ -825,7 +830,7 @@ async function verifyAndroidRun({
     waitedMs: verification?.waitedMs,
     bundleId: androidPackage,
     serial,
-    devClientUrl: scheme ? androidDevClientUrl(scheme, metroPort ?? DEFAULT_METRO_PORT) : null,
+    devClientUrl: scheme ? androidDevClientUrl(scheme, metroPort ?? DEFAULT_METRO_PORT, physical) : null,
     mode: isExpo ? MODE_EXPO : MODE_BARE,
   }))
     phase('', chalk.yellow(line));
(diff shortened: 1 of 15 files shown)


packages/stim-cli/src/commands/android.ts
```diff
@@ -1626,69 +1666,82 @@ export async function runAndroid(options: RunAndroidOptions = {} as RunAndroidOp
...
-  const limits = getLimits();
-  const capacity = checkCapacity({
-    platform: PLATFORM,
-    project,
-    max: limits.maxDevices,
-  });
-  if (capacity) return fail(capacity.code, capacity.message, capacity.remedy);
-
   const emuLog = emulatorLogFile(root);
+  const limits = getLimits();
```
janicduplessis, line 1687: `getLimits()` is hoisted out of the emulator branch on purpose. The device-capacity check below it is emulator-only -- `maxDevices` counts devices Stim owns, and a physical device is not one -- but `limits.maxBuilds` gates the build slot further down, and the build runs locally whether the target is a phone or an emulator. Leaving the call inside the else branch left `limits` out of scope at the build-slot site.

PR #188, opened 2026-09-01
Ignore the paths a native build never reads

### LINKED ISSUE #187: Ship the build-irrelevant fingerprint ignores as defaults
## What happens

Every project that wants a fingerprint to survive a second checkout writes the
same `.fingerprintignore`. The entries are not project knowledge. They are the
same five paths in every React Native app:

```
ios/**/Package.resolved              # written by Xcode on first SPM resolve
android/local.properties             # generated, carries an absolute sdk.dir
android/.idea/**                     # appears once someone opens Android Studio
**/*.map                             # source maps, never a native input
**/node_modules/**/lib/**            # built JS in third-party packages
```

The first three are correctness: each is absent from a fresh checkout and
present in a working one, so the two fingerprint differently and the cache
cannot hit across checkouts. The last two are compute: in one real app they are
about 26% of fingerprint time, because one dependency's `lib/` alone holds 3137
files.

`@expo/fingerprint` already defaults to ignoring 37 paths, including
`**/ios/Pods/**/*` and `**/android/build/**/*`, but none of the five above.

## Why Stim should carry them

Stim computes the fingerprint that keys its own build cache, so it can decide
what a native build reads. A project that keeps its own `.fingerprintignore`
still works, and the file stays the place for genuinely project-specific
entries.

None of the five can change a native build:

- `Package.resolved` records resolved SPM versions; the pods and the checked-in
  project define what compiles.
- `local.properties` holds the machine's SDK path.
- `.idea/**` is IDE state.
- A `.map` is a debugging artifact for JavaScript.
- `lib/` in a dependency is its built JavaScript. Autolinking reads native
  sources: `cpp/`, `apple/`, `android/`, the podspec. Config plugins live under
  `plugin/build/`, not `lib/`.

## Proposal

Pass these as `ignorePaths` from `fingerprintProject`, which every Stim
fingerprint already goes through, so `doctor`'s parity check and the build
cache key agree. Document them in the guide next to the `.fingerprintignore`
advice.

janicduplessis:
Closes #187.

## Description

A fresh checkout and a working checkout of the same commit fingerprint differently, so the build cache cannot hit across them — the parity failure `doctor` reports. Two of the causes are the same in every project: `android/local.properties`, which is generated and carries this machine's `sdk.dir`, and `android/.idea`, which appears the first time someone opens Android Studio. `@expo/fingerprint` ignores 37 paths by default; neither of these.

## Solution

`fingerprintProject` passes both as `ignorePaths`. It is the one place every Stim fingerprint goes through — the build cache key and `doctor`'s parity check — so the two cannot disagree. `ignorePaths` merges with the built-in defaults and with the project's `.fingerprintignore`, which still applies on top and remains the place for anything only that project can judge.

The list is deliberately just these two. An earlier version also ignored `**/*.map`, `**/node_modules/**/lib/**` and `**/ios/**/Package.resolved`; review found a real counterexample for each. An Expo config plugin can ship under a dependency's `lib/` (`react-native-vision-camera` resolves `app.plugin.js` to `lib/commonjs/…`), and fingerprint records every loaded plugin as a source, so on a CNG project that can hide a plugin-only dependency change. `.map` is also a bundled-asset extension — offline map data lives under `android/app/src/main/assets`. `Package.resolved` pins the revision an SPM version range resolved to, so ignoring it can serve a stale build. A test now asserts none of the three comes back.

## Two consequences worth knowing

**Hashes diverge from Expo CLI.** Stim computes this fingerprint itself, so on a machine that has these files, Stim's hash and the hash Expo CLI hands `@stim-cli/expo-build-cache` differ, and each side fills its own cache entry. Both provider docs now say so, and a project that wants them aligned lists the same paths in its `.fingerprintignore`.

**Existing cache entries go cold once.** The default fingerprint changes, so entries under `~/.stim/build-cache` from before the upgrade are unreachable: one cold build per project, platform and variant, and the old entries linger until `stim gc`.

## Test plan

- `pnpm test` — 2840 pass, including: the defaults reach `@expo/fingerprint` alongside platform scoping; the three rejected paths stay out of the list; and the guide names every path the list holds, so prose and code cannot drift
- `format:check`, `lint`, `build`, `typecheck` — clean
- Verified in `@expo/fingerprint@0.20.10` (`Options.js:collectIgnorePathsAsync`) that `ignorePaths` is concatenated with `DEFAULT_IGNORE_PATHS` and the project's `.fingerprintignore` rather than replacing either, and that a `!pattern` line can still un-ignore one of these

### DIFF
--- a/packages/expo-build-cache/README.md
+++ b/packages/expo-build-cache/README.md
@@ -2,6 +2,11 @@
 
 A local Expo build-cache provider that shares `.app` and `.apk` artifacts across
 worktrees and direct Expo builds.
+Stim and Expo CLI compute the fingerprint that keys this cache separately, and
+Stim ignores two machine-local Android paths (`android/local.properties`,
+`android/.idea`) that Expo CLI does not. On a machine where those exist, the
+two hashes differ and each side fills its own entry. List them in the project's
+`.fingerprintignore` to bring the hashes back together.
 
 If Stim is not installed globally, replace `stim` with `npx stim-cli`.
 
--- a/packages/stim-cli/src/__tests__/build-cache.test.ts
+++ b/packages/stim-cli/src/__tests__/build-cache.test.ts
@@ -27,6 +27,7 @@ import {
   entryDir,
   fingerprintDiffRecord,
   fingerprintDiffSuffix,
+  DEFAULT_FINGERPRINT_IGNORES,
   fingerprintProject,
   prepareProviderDownloadDir,
   providerDownloadPath,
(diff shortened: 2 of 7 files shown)


janicduplessis:
Not ready to merge. CI is green, but a fresh review found a blocker in the approach, not the code.

`@stim-cli/expo-build-cache` receives `fingerprintHash` from Expo CLI and looks up `buildCacheKey(platform, fingerprintHash, runOptions)` in the same `buildCacheRoot()` the CLI uses. Expo computes that hash without Stim's `ignorePaths`, so once the CLI applies them the two keys can never agree, and the sharing that `packages/expo-build-cache/README.md` and `website/docs/cache-packages.md` promise between Stim builds and direct Expo builds stops working. CLAUDE.md's cache contract asks for the opposite: keep the package's cache key aligned with the CLI.

Two entries in the list are also wrong as justified:

- `**/node_modules/**/lib/**` drops real Expo config plugins. `react-native-vision-camera@4.7.2/app.plugin.js` is `module.exports = require('./lib/commonjs/expo-plugin/withVisionCamera')`, and fingerprint records every loaded plugin as a file source. On a CNG project, where ios/ and android/ are ignored anyway, that can be the only signal a plugin-only dependency changed. The comment claiming plugins live under `plugin/build/` is factually wrong.
- `**/*.map` is not only source maps. Offline map data (mapsforge/VTM `*.map` under `android/app/src/main/assets/`) is a real APK input.

Verified and fine: `ignorePaths` merges with the 37 built-in defaults and the project's `.fingerprintignore` rather than replacing them, and a project can still un-ignore one with a `!pattern` line.

Holding this until the interop question is decided.

PR #212, opened 2026-09-02
feat: name the sandbox allowance in doctor, apply it with --fix

### LINKED ISSUE #199: doctor should name the sandbox allowance, and --fix should apply it
## Why

`guide errors` now names what a sandbox blocks, and the skill sends an agent
there. Both still leave the user to translate that into their harness's
config by hand, and doctor -- the command whose whole job is "your machine is
not set up for fast builds" -- says nothing, even though it can see both the
harness and the setting.

## What doctor can determine

- **The harness.** Claude Code exports `CLAUDECODE` and `CLAUDE_CODE_ENTRYPOINT`.
  Codex exports `CODEX_*` and keeps `~/.codex/config.toml`.
- **Whether it is already handled.** Under Claude Code the three keys either
  appear in a settings file or they do not.

## What to report

A finding when a sandboxing harness is present and the allowance is missing,
carrying the patch verbatim:

    sandbox.filesystem.allowWrite     ["~/.stim"]
    sandbox.network.allowMachLookup   ["com.apple.coresimulator.*"]
    sandbox.network.allowLocalBinding true

## `--fix`

`doctor --fix` writes that patch, and only that patch, into
`.claude/settings.local.json` -- the personal, gitignored file, never the
committed project one -- merging with what is there rather than replacing it.

It refuses under Codex, and says why: the sandbox is one enum
(`sandbox_mode = "workspace-write"`), there is no per-path allowance, and the
only value that clears `STIM_HOME` is `danger-full-access`. A tool should not
switch a user's sandbox off wholesale, so Codex gets the explanation and no
write.

With no harness detected, `--fix` reports that there is nothing to apply.

## Worth being deliberate about

Stim would be writing another tool's config file, so `--fix` stays opt-in
rather than something `doctor` does on its own, and it encodes a schema Stim
does not own, which will drift. The printed finding never drifts, so it is the
part that always runs; the write is the convenience.

janicduplessis:
## Description

`guide errors` already names what a sandbox blocks and the skill routes an
agent there, but both leave the reader to translate that into their harness's
config by hand. `doctor` -- the command whose job is "your machine is not set
up for fast builds" -- said nothing, even though it can see both the harness
and the setting.

Two things a sandboxed harness makes invisible are in scope here: the harness
identity (Claude Code exports `CLAUDECODE` / `CLAUDE_CODE_ENTRYPOINT`; Codex
exports `CODEX_*` and keeps `~/.codex/config.toml`) and, under Claude Code,
whether the three allowance keys are already satisfied in a settings file.

Blast radius: `doctor.ts`, `commands/doctor.ts`, `commands/guide.ts`, a new
`src/sandbox.ts`, the guide contract tests, and the website command
reference. No other command changes, and no runtime path changes.

## Solution

Two clearly separated parts, matching #199.

**(a) The finding. Always runs, no flag.** `runDoctor` gains a sandbox check.
When a sandboxing harness is detected and the allowance is missing, it reports
a finding carrying the patch verbatim. Under Claude Code, "already handled" is
decided by a pure function (`sandboxAllowanceSatisfied`) from the parsed
contents of `.claude/settings.local.json`, `.claude/settings.json`, and
`~/.claude/settings.json`. Satisfied is by value, not by key presence: an
`allowWrite` array that exists without `~/.stim` is still a missing allowance,
which keeps the finding and `--fix` in agreement (the finding disappears
exactly when `--fix` would be a no-op). Under Codex the finding is a `note`
that explains the single-enum sandbox and recommends nothing automatic.

**(b) `--fix`. Opt-in.** Under Claude Code it merges exactly that patch into
`.claude/settings.local.json`, creating it when absent, and prints what it
wrote to stderr. It refuses under Codex with the explanation, and reports
there is nothing to apply when no harness is detected. It never touches the
committed `.claude/settings.json`.

The whole file computation is one pure function, `mergeSandboxAllowance(source)`,
that takes the file's text (or `null`) and returns either a refusal or the
merged JSON. It deep-merges, appends to arrays without duplicating, keeps every
other key, and refuses rather than clobbers when the file does not parse or
when a path it needs holds a non-object/non-array/non-boolean. The I/O wrapper
around it does an atomic write (tmp + rename), the pattern `config.ts` and
`paths.ts` already use.

Deliberate trade-offs:

- The patch has one source. `SANDBOX_ALLOWANCE` in `sandbox.ts` produces both
  the printed lines and the merge; a contract test asserts the guide topic and
  the finding print the same three lines.
- `--fix` stays opt-in because Stim is writing another tool's config file and
  encoding a schema it does not own, which will drift. The printed finding does
  not drift, so that is the part that always runs.
- The Codex refusal names `danger-full-access` as the only value that clears
  `STIM_HOME` but recommends nothing automatic: switching a user's sandbox off
  wholesale is the user's call.
- AGENTS.md's "never mutates project setup" sentence now states the one
  exception in the present tense, and keeps "do not add either" for init/setup.
- `SKILL.md` is unchanged. The normal workflow, the safety rules, and topic
  routing are all the same -- the skill already says a sandbox exists and sends
  the reader to `stim guide errors` -- so invariant 1 says leave it alone. It
  also has 6 words of headroom under the 1,200-word cap.

## Test plan

New unit tests for the pure merge, against exactly the cases #199 asks for:
absent file, existing unrelated keys, existing partial allowance, existing full
allowance (no-op), malformed JSON (refuse, do not overwrite), plus shapes it
would have to clobber. Plus detection precedence, value-not-key satisfaction,
the atomic write leaving the committed file alone, the Codex refusal, and the
no-harness report.

New contract tests pin the flag and the finding to the guide: `doctor.ts`'s
`--json` / `--fix` are added to the "flags the guide advertises are the flags
the commands define" map, and a new test asserts the `errors` topic carries the
same three lines the finding carries, names `CLAUDECODE`, `CODEX_`,
`~/.codex/config.toml` and `.claude/settings.local.json`, says it never touches
the committed file, and that the skill does not carry `--fix`.

```text
pnpm run format:check   All matched files use the correct format.
pnpm run lint           clean
pnpm run build          Build complete
pnpm run typecheck      clean
pnpm test               78 files, 2961 tests passed
pnpm run knip           exit 0
```

`test:e2e` was not run: no e2e case touches `doctor`, and no end-to-end
workflow changed.

### The built CLI, in a scratch project

Scratch project with a committed `.claude/settings.json` holding an unrelated
key, `HOME` pointed at an empty directory so no real user settings leak in.

**1. `doctor` under `CLAUDECODE=1` shows the finding**

```text
costs time  This Claude Code session has no sandbox allowance for Stim
  A sandboxed shell blocks three things Stim needs, and none of the failures names the sandbox: writes to STIM_HOME (~/.stim unless set), the CoreSimulator service simctl reaches over XPC, and the adb server socket on tcp:5037. None of .claude/settings.local.json, .claude/settings.json, or ~/.claude/settings.json carries all three keys:

    sandbox.filesystem.allowWrite     ["~/.stim"]
    sandbox.network.allowMachLookup   ["com.apple.coresimulator.*"]
    sandbox.network.allowLocalBinding true

  The allowance changes nothing when the sandbox is off, so applying it costs nothing and saves three failures that read as a broken machine.
  -> `stim doctor --fix` merges exactly those keys into .claude/settings.local.json, the personal gitignored file, leaving every other key alone. See `stim guide errors`.

1 finding(s). Fix relevant "costs time" findings before copying the main checkout into a native worktree.
```

**2. `doctor --fix` writes the file (stderr)**

```text
Wrote the sandbox allowance into /private/tmp/stim199-scratch-24p6/.claude/settings.local.json (sandbox.filesystem.allowWrite, sandbox.network.allowMachLookup, sandbox.network.allowLocalBinding). Every other key in that file is untouched, and the committed .claude/settings.json is never written. Restart the harness session for it to take effect.
```

```json
// .claude/settings.local.json, created
{
  "sandbox": {
    "filesystem": { "allowWrite": ["~/.stim"] },
    "network": { "allowMachLookup": ["com.apple.coresimulator.*"], "allowLocalBinding": true }
  }
}
// .claude/settings.json, unchanged
{ "permissions": { "allow": ["Bash(ls:*)"] } }
```

**3. A second `doctor` shows it handled**

```text
Nothing to flag.
Checked: main-checkout dependencies, CocoaPods, native warm state and local upstream; dev client, ccache, Metro cacheStores, compilation cache, build cache provider, EAS session, SimSlim profile, the sandbox allowance of a detected agent harness and, on a checkout without installed dependencies, fingerprint parity.
```

**4. A Codex-env run refuses (`CODEX_SANDBOX=seatbelt`, Claude vars unset)**

```text
Codex has one sandbox setting rather than a per-path allowance: sandbox_mode is read-only, workspace-write, or danger-full-access. workspace-write still refuses STIM_HOME, so danger-full-access is the only value that clears it. Turning a sandbox off wholesale is the user's call, not a tool's, so Stim writes nothing here.

note  This is a Codex session, whose sandbox has no per-path allowance
  A sandboxed shell blocks three things Stim needs, and none of the failures names the sandbox: writes to STIM_HOME (~/.stim unless set), the CoreSimulator service simctl reaches over XPC, and the adb server socket on tcp:5037. Codex has one sandbox setting rather than a per-path allowance: sandbox_mode is read-only, workspace-write, or danger-full-access. workspace-write still refuses STIM_HOME, so danger-full-access is the only value that clears it. Turning a sandbox off wholesale is the user's call, not a tool's, so Stim writes nothing here.
  -> Decide before the session starts whether this one runs Stim, and if it does, run Codex with danger-full-access. `stim doctor --fix` refuses here.
```

No `.claude/` directory is created on that run.

**5. No harness detected**

```text
No sandboxing harness is in this environment, so there is nothing to apply. The allowance only means something to a harness that sandboxes shell commands.
```

**6. Deep merge into an existing personal file, and idempotence**

Starting file had `permissions.allow`, `sandbox.enabled`, an unrelated
`allowWrite` entry, and `sandbox.network.allowUnixSockets`:

```json
{
  "permissions": { "allow": ["Bash(git status:*)"] },
  "sandbox": {
    "enabled": true,
    "filesystem": { "allowWrite": ["~/.other", "~/.stim"] },
    "network": {
      "allowUnixSockets": ["/tmp/a.sock"],
      "allowMachLookup": ["com.apple.coresimulator.*"],
      "allowLocalBinding": true
    }
  }
}
```

A second `--fix`: `... already carries the sandbox allowance. Nothing to write.`

**7. A malformed personal file is refused, not overwritten**

```text
Refusing to write /private/tmp/stim199-merge-JQKU/.claude/settings.local.json: it is not valid JSON: Expected property name or '}' in JSON at position 2 (line 1 column 3). Nothing was changed. Repair the file, or add the three keys by hand.
```

The file still reads `{ oops` afterwards.

**8. `--fix --json` keeps the stdout contract (invariant 7)**

The message goes to stderr; stdout carries exactly one payload, which gains a
`fix` object only when `--fix` ran:

```json
{
  "project": "/private/tmp/stim199-merge-JQKU",
  "findings": [],
  "fix": { "harness": "claude", "applied": true, "path": "...", "message": "..." }
}
```

Invariant 9 does not apply: no `git`, `simctl`, `adb`, `avdmanager`,
`xcodebuild` or Gradle argument list changed.

Fixes #199

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -23,7 +23,10 @@ need custom behavior.
 
 Runtime state belongs under `$STIM_HOME/workspaces/`, not in the project
 tree. Stim has no init step and never mutates project setup; do not add
-either. `doctor` reports setup that requires project judgment.
+either. `doctor` reports setup that requires project judgment. The single
+exception is `doctor --fix`, which writes the sandbox allowance into the
+personal, gitignored harness settings file (`.claude/settings.local.json`) and
+nothing else; it never writes a committed file.
 
 ## Development
 
--- a/packages/stim-cli/src/__tests__/doctor.test.ts
+++ b/packages/stim-cli/src/__tests__/doctor.test.ts
@@ -19,9 +19,11 @@ import {
   detectXcodeMajor,
   parseXcodeMajor,
   checkConcurrency,
+  checkSandboxAllowance,
 } from '../doctor.ts';
 import { resetExecutor, setExecutor } from '../exec.ts';
 import type { EasAuthResult } from '../engine/remote-cache.ts';
+import { sandboxAllowanceLines } from '../sandbox.ts';
 import assert from 'node:assert';
 
 test('checkMainCheckout reports missing dependencies, Pods, and native output', () => {
@@ -916,6 +918,67 @@ test.each([
 
     expect(findings.some((finding) => finding.title === 'The remote proxy credentials are missing')).toBe(true);
     expect(findings.some((finding) => finding.title === 'This project uses a remote proxy')).toBe(false);
+  } finally {
+    rmSync(project, { recursive: true, force: true });
+  }
+});
+
+test('checkSandboxAllowance stays silent with no harness and with the allowance in place', () => {
+  expect(checkSandboxAllowance({ harness: null })).toBe(null);
+  expect(checkSandboxAllowance({ harness: 'claude', satisfied: true })).toBe(null);
+});
+
+test('the Claude Code finding carries the patch verbatim and points at --fix', () => {
+  const found = checkSandboxAllowance({ harness: 'claude', satisfied: false });
+  assert(found);
+  expect(found.level).toBe('cost');
+  for (const line of sandboxAllowanceLines()) expect(found.detail).toContain(line);
+  expect(found.detail).toContain('.claude/settings.local.json');
+  expect(found.fix).toMatch(/doctor --fix/);
+});
+
+test('the Codex finding explains the single enum and recommends nothing automatic', () => {
+  const found = checkSandboxAllowance({ harness: 'codex' });
+  assert(found);
+  expect(found.level).toBe('note');
+  expect(found.detail).toMatch(/sandbox_mode/);
+  expect(found.detail).toMatch(/danger-full-access is the only value/);
+  expect(found.detail).not.toContain('sandbox.filesystem.allowWrite');
+  expect(found.fix).toMatch(/refuses/);
+});
+
+test('runDoctor reports the sandbox allowance from the project settings files', () => {
+  const project = mkdtempSync(join(tmpdir(), 'stim-doc-sandbox-'));
+  try {
+    writeFileSync(join(project, 'package.json'), JSON.stringify({ name: 'x' }));
+    const missing = runDoctor(project, {
+      concurrency: () => ({ maxBuilds: 0, maxDevices: 0 }),
+      harness: () => 'claude',
+      claudeSettings: [],
+    });
+    expect(missing.some((f) => /sandbox allowance/i.test(f.title))).toBe(true);
+
+    mkdirSync(join(project, '.claude'), { recursive: true });
+    writeFileSync(
+      join(project, '.claude', 'settings.local.json'),
+      JSON.stringify({
+        sandbox: {
+          filesystem: { allowWrite: ['~/.stim'] },
+          network: { allowMachLookup: ['com.apple.coresimulator.*'], allowLocalBinding: true },
+        },
+      }),
+    );
+    const handled = runDoctor(project, {
+      concurrency: () => ({ maxBuilds: 0, maxDevices: 0 }),
+      harness: () => 'claude',
+    });
+    expect(handled.some((f) => /sandbox/i.test(f.title))).toBe(false);
+
+    const none = runDoctor(project, {
+      concurrency: () => ({ maxBuilds: 0, maxDevices: 0 }),
+      harness: () => null,
+    });
+    expect(none.some((f) => /sandbox/i.test(f.title))).toBe(false);
   } finally {
     rmSync(project, { recursive: true, force: true });
   }
--- a/packages/stim-cli/src/__tests__/guide.test.ts
+++ b/packages/stim-cli/src/__tests__/guide.test.ts
@@ -5,6 +5,8 @@ import { DEFAULT_FINGERPRINT_IGNORES } from '../build-cache.ts';
 import { topicNames, renderTopic, renderIndex } from '../commands/guide.ts';
 import { ANDROID_AVD_CONFIG_HELP } from '../settings.ts';
 import { CONSOLE_ENV, deviceConsoleArgs } from '../collector/ios-device.ts';
+import { checkSandboxAllowance } from '../doctor.ts';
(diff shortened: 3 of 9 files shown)


janicduplessis:
Duplicate: PR #202 landed the same feature on main (6640771) while this was in review. Closing unmerged; nothing here is needed on top of #202 except website docs for the flag, tracked separately.

PR #238, opened 2026-09-02
fix: pool refusal lists each unhealthy device's own reason

### LINKED ISSUE #235: Pool with no healthy candidates falls back to the resolver's 'Several devices are connected' message
## Problem

With two or more cabled iPhones that are all unhealthy (Developer Mode off, unpaired), an id-less `ios --device` has no pool candidates and falls back to `resolveIosPhysicalDevice(null, ...)`, whose several-cabled branch says "Several devices are connected ... Name the one to build for" instead of the Developer Mode or pairing remedy. Naming one device then produces the right refusal, so it is a detour, not a dead end. Android has the same shape when every physical serial is an emulator (rare).

## Evidence

Fresh review of PR #234 (2026-09-02), by reading `engine/device-pool.ts` `noCandidates` and `engine/ios-device.ts` ~:146 (`cabled.length > 1` branch). Not reproduced on hardware (one phone connected).

## Cause

`noCandidates` reuses the single-device resolver for its message, and that resolver checks the count of cabled devices before health.

## Fix idea

When the pool has cabled devices but none healthy, report the health refusal for each (the resolver's own unhealthy message per device) instead of the count message; keep the count message only when every cabled device is healthy and an id is genuinely needed (which the pool now makes unreachable). One unit test with two unhealthy devices. Optional in the same PR: a unit test for `memoizeEmulatorProbe`.

janicduplessis:
## Description

Two cabled iPhones that are both unhealthy (Developer Mode off, unpaired) make an id-less `ios --device` take a detour: the pool has no healthy candidate, so it falls back to the single-device resolver, and that resolver checks the cabled *count* before health. It answers "Several devices are connected ... Name the one to build for" instead of the Developer Mode or pairing remedy. Naming one device then produces the right refusal, so it's a detour, not a dead end -- but it hides the actual problem from an agent that only ever runs id-less. Android has the same shape when every physical serial is an emulator.

The same `noCandidates` mistake exists at three call sites: `runIos` (`commands/ios.ts`), `pooledAndroidDevice` (`commands/android.ts`), and `poolDevice` (`commands/device.ts`, behind an id-less `stim device lock`).

## Solution

`iosPoolNoCandidatesRefusal` (`engine/ios-device.ts`) and `androidPoolNoCandidatesRefusal` (`sim/android.ts`) replace the resolver call at all three call sites. When the pool has cabled/physical devices but none is a valid candidate, they build the refusal from each device's own reason instead of the ambiguous-count branch -- reusing the resolver's exact per-device wording (`unhealthy()` on iOS, `emulatorRefusal()` on Android), not new text.

The old count-message branch is still there and still reachable, just narrower: it now only fires when the pool has zero cabled/physical devices connected at all.

## Test plan

Each repro test below was red against the pre-fix code (asserted the buggy "Several devices are connected" / "Several physical devices are connected" text) before the corresponding call site was fixed, then green after.

- `ios-command.test.ts` / `android-command.test.ts`: end-to-end repro for `runIos` / `pooledAndroidDevice` -- two cabled phones (one Developer-Mode-off, one unpaired) / two emulator-probed serials, `--device` with no id now refuses `STIM_NO_DEVICE` naming each device's own reason.
- `device-command.test.ts`: same repro for `poolDevice`, via an id-less `stim device lock ios` / `stim device lock android`.
- `engine-ios-device.test.ts` / `sim-android.test.ts`: unit tests for the two new pure functions, including their fallback to the existing resolver when nothing is cabled/connected.
- `sim-android.test.ts`: new unit test for `memoizeEmulatorProbe` -- one probe per serial across repeated calls.
- Checked `guide.ts` for a pinned "Several devices" sentence -- none exists, so no guide update was needed.
- Full battery from the repo root: `pnpm run format:check`, `pnpm run lint`, `pnpm run build`, `pnpm run typecheck`, `pnpm test` (82 files, 3166 tests) all pass.

Fixes #235

### DIFF
--- a/packages/stim-cli/src/__tests__/android-command.test.ts
+++ b/packages/stim-cli/src/__tests__/android-command.test.ts
@@ -4101,6 +4101,17 @@ describe('--device with no serial: the pool', () => {
     expect(result.error?.code).toBe(NO_DEVICE);
   });
 
+  test('two emulator-only serials refuse with each one own reason, not the count message', async () => {
+    const h = pool([FIRST, SECOND], { isEmulatorDevice: () => true });
+    const result = await h.run();
+    expect(result.ok).toBe(false);
+    expect(result.error?.code).toBe(NO_DEVICE);
+    expect(result.error?.message).toMatch(new RegExp(`${FIRST} is an emulator, not a physical device`));
+    expect(result.error?.message).toMatch(new RegExp(`${SECOND} is an emulator, not a physical device`));
+    expect(result.error?.message).not.toMatch(/Several physical devices are connected/);
+    expect(result.error?.remedy).toMatch(/without --device/);
+  });
+
   test('a leased device that is not connected refuses, naming it and the way out', async () => {
     takeLease({ root, platform: 'android', id: 'GONE-SERIAL', kind: 'declared' });
     const h = pool([FIRST]);
--- a/packages/stim-cli/src/__tests__/device-command.test.ts
+++ b/packages/stim-cli/src/__tests__/device-command.test.ts
@@ -277,6 +277,53 @@ describe('stim device lock', () => {
     expect(facts.deviceName).toBe('SM-G996W');
   });
 
+  test('with no id and two unhealthy cabled iOS phones, the refusal names each one own reason', async () => {
+    const SECOND = '00008120-000A11223C44201E';
+    const h = harness({
+      listIosDevices: () => [
+        {
+          udid: PHONE,
+          name: 'Old iPhone',
+          bootState: 'booted',
+          developerModeStatus: 'disabled',
+          pairingState: 'paired',
+          transportType: 'wired',
+        },
+        {
+          udid: SECOND,
+          name: 'Second',
+          bootState: 'booted',
+          developerModeStatus: 'enabled',
+          pairingState: 'unpaired',
+          transportType: 'wired',
+        },
+      ],
+    });
+    const failure = await runLock('ios', undefined, {}, h.deps);
+    assert('code' in failure);
+    expect(failure.code).toBe('STIM_NO_DEVICE');
+    expect(failure.message).toMatch(new RegExp(`${PHONE} \\(Old iPhone\\) has Developer Mode disabled`));
+    expect(failure.message).toMatch(new RegExp(`${SECOND} \\(Second\\) is connected but unpaired`));
+    expect(failure.message).not.toMatch(/Several devices are connected/);
+  });
+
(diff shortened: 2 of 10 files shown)


janicduplessis (commented):
Reviewed with fresh eyes against issue #235 and AGENTS.md. (GitHub blocks self-approval since all agents share one account, so posting as a comment: this is an approve-quality review, no blocking findings.)

**Correctness.** `iosPoolNoCandidatesRefusal`/`androidPoolNoCandidatesRefusal` are only invoked from each `noCandidates` callback, which `selectFromPool` calls only when `list()` (i.e. `iosPoolCandidates`/`androidPoolCandidates`, which already filter to wired+healthy / physical-non-emulator) is empty. So the "mixed healthy/unhealthy cabled" case can't actually reach these functions today -- a healthy candidate means `noCandidates` never fires. Single-unhealthy-device and wireless-only-device cases fall through unchanged to the existing resolver, which already handled them correctly. The fix matches the issue's proposed shape exactly: per-device reasons joined, falling back to the resolver only when nothing is cabled/physical.

**Duplication/architecture.** Both new functions re-derive the cabled/physical filter that the fallback resolver also computes, but they're pure, side-effect-free, and independently correct, consistent with the "pure parsing and decision logic" rule. `memoizeEmulatorProbe` (pre-existing, already wired at commands/android.ts:1399) means the new Android refusal path doesn't cause extra `adb`/`getprop` probing.

**Tests.** The new tests assert both presence of each device's own reason and absence of the old count message/text, which actually pins the fix rather than just exercising the path. `memoizeEmulatorProbe` test correctly proves one-probe-per-serial caching.

**Docs.** Confirmed (as the PR description also notes) there's no "Several devices are connected" text pinned in guide.ts/README/SKILL.md, so no doc update was needed.

**Comment policy.** No new comments added; consistent with the file's existing style.

**Checks run in a scratch clone of `fix/pool-no-healthy-candidates`:** `pnpm install`, `format:check`, `lint`, `build`, `typecheck` all clean; `pnpm test` -- 82 files, 3164 tests, all passing.

Minor, non-blocking observations: the Android refusal only ever reports "emulator, not physical" reasons -- a physical-only pool that's all-emulators *and* also has an unrelated adb-`unauthorized` entry in `unhealthy` would drop that second reason (pre-existing single-reason limitation elsewhere too, not a regression). Also `iosPoolNoCandidatesRefusal`'s `problems.every(...)` check re-derives an invariant the call site already guarantees, which is harmless defensiveness rather than a bug.

No actionable findings; this looks ready to mark ready for merge.

Issue #260, opened 2026-09-02
Run and lifecycle output: one line vocabulary, no OS log noise on a verified launch, facts not explanations


janicduplessis:
## Problem

The stderr a run prints mixes three formats and repeats itself, and a passing simulator launch ends with a dozen `launch err` lines of OS noise. Agents read every line; the maintainer's example: after a cache hit, the `install     skipped; ... already holds this app (45.6s)` line reads as a problem. Real transcripts from the rc.8 gate (2026-09-02, `stim-bench-coordinator/release-rc8/qa-loop-expo-ios.log`, `qa-loop-expo-android.log`, `manual-ios.md`, `/tmp/b-plain.err`):

```
Created owned sim stim-e2e-1 (FD36AE9A-A23F-4935-BA23-21CC097C1EDC)
  device      stim-e2e-1 (FD36..) created (36.2s)
  fingerprint dcbd8d.. miss (6.2s)
  prebuild    ios/ absent -> generated (5.2s)
pods        still running (30s)
pods        still running (1m30s): [Expo] Ensuring required slices in ExpoModulesJSI.xcframework
  pods        ios/Podfile exists but no pods are installed -> installed with `pod install` (1m37s)
  fingerprint dcbd8d.. -> 6564e2.. after prebuild + pod install; storing under the new key, which is the one the next run looks up
  build       compiling Debug with xcodebuild
compilation cache on for this build: CAS at /var/folders/.../compilation-cache (Stim ...
build       still running (30s):     /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctool...
compilation cache 0/171 hits (0%)
  build       ok (51.8s)
  device      stim-e2e-1 (FD36..) booted (3.6s)
  install     -> stim-e2e-1 (FD36..) (4.1s)
  launch      com.janic.app (2s)
  verify      bundle loaded, process alive, stable for 3s -- the first screen may still be rendering (4.6s total)
  launch err  attention client lost event tag is not a number: (null)
  launch err  attention client com.apple.SpringBoard.HomeGrabberView-6 cannot reset; error:Error Domain=NSPOSIXErrorDomain Code=57 ...
  launch err  [S:2a] Error received: Connection invalidated.
  launch err  *** Assertion failure (Identifier: catch-all) : +[FCExcerptURLProtocol setupWithArticleDatabase:] ...
  launch err  amsengagementd.AppURLSession: [075DFF65_34] Object decoding failed. ...
```

[...]

PR #266, opened 2026-09-03
fix: anchor the iOS simulator log predicate to the app's executable path

### LINKED ISSUE #264: Simulator log collector predicate matches every process whose path contains the app name
## Problem

The iOS simulator collector filters `log stream` with `processImagePath CONTAINS[c] "<app basename>"` (`collector/ios.ts` ~:145-156). For an app named `app` (the native e2e fixture's `app.app`, bundle `com.janic.app`) that admits every process on the simulator whose image path contains `app`, so SpringBoard, amsengagementd, JetEngine, and the FC*URLProtocol assertions land in the workspace's device log and, before #260, in the run summary as `launch err` lines. Any short app name (`app`, `mobile`, `ios`) leaks the same way; a name like `Trailhead` does not.

## Evidence

rc.8 gate loop transcript, `stim-bench-coordinator/release-rc8/qa-loop-expo-ios.log` lines 60-64 and 90-101: SpringBoard, `[S:2a] Error received: Connection invalidated`, `amsengagementd.AppURLSession` records under a verified launch of `com.janic.app`. `~/.stim/workspaces/issue-245-measure--*/logs/device.ndjson` for Trailhead: all error-level records carry `proc: "Trailhead"`, none foreign. Found during the review of PR #261 (2026-09-02).

## Cause

`CONTAINS[c]` on the basename instead of an anchored match on the app's executable path.

## Fix idea

Anchor the predicate to the installed app's executable: `processImagePath ENDSWITH "/<AppName>.app/<CFBundleExecutable>"` (read CFBundleExecutable from the built app's Info.plist, which `ios.ts` ~:477 already parses), or `process == "<CFBundleExecutable>"` plus the path anchor. Keep the app's own extensions out unless a design says otherwise. Unit-test the predicate string; real-tool check once on a simulator with a short-named fixture (the native e2e `app` fixture) showing SpringBoard records gone from device.ndjson while the app's own records remain. Then the count line from #260 measures only the app's process, as its guide text says.

janicduplessis:
## Description

The device-log collector filtered `xcrun simctl spawn <udid> log stream` with `processImagePath CONTAINS[c] "<app basename>"` (`collector/ios.ts` `logStreamArgs`). That admits any process whose image path merely *contains* the app name — for a short or common app name (the native e2e fixture's `app`, or names like `mobile`/`ios`) this pulled SpringBoard, `amsengagementd`, and other unrelated system processes into `device.ndjson`, and (before #260) into the run's error summary.

PR #261 also touched `commands/ios.ts` around the `replaceCollector` call site and has since merged; this branch is rebased onto it. My hunk there stays small — one new local (`appExecutable`), passed to the simulator's `replaceCollector` call only (the physical-device call site doesn't need it; hardware streams the devicectl console, not `log stream`).

## Solution

Anchor the predicate to the app's own executable path instead: `processImagePath ENDSWITH "/<AppName>.app/<CFBundleExecutable>"`, falling back to the `.app` basename when `CFBundleExecutable` can't be read (that's what Xcode uses as the executable name by default anyway).

`CFBundleExecutable` comes from a new `readBundleExecutable` in `engine/xcode.ts`, parsed from the built app's `Info.plist` the same way `readBundleId` already reads `CFBundleIdentifier`. `commands/ios.ts` reads it once and threads it through `replaceCollector` -> the collector subprocess (`--app-executable`) -> `startIosLogStream`. `procFromImagePath` needed no change: it already takes the last path segment, which is the executable name. When `CFBundleExecutable` can't be read, `commands/ios.ts` now also emits a dim stderr note naming the app path, so an unexpectedly empty device log is diagnosable instead of silently falling back.

## Test plan

- Unit tests: `logStreamArgs`/`startIosLogStream` (`collector-parsers.test.ts`) cover the anchored predicate, the CFBundleExecutable-vs-basename case, the fallback, and directly assert that a short app name (`app`) no longer matches `/SpringBoard.app/SpringBoard` or `/usr/libexec/amsengagementd` while it does match the app's own installed path. `readBundleExecutable`/`parseBundleExecutable` (`engine-xcode.test.ts`) mirror the existing `readBundleId` coverage. `ios-command.test.ts` and `collector-run.test.ts` cover the plumbing (`--app-executable` on the spawned collector argv and its parsing), the diagnostic note when `CFBundleExecutable` can't be read, and the real-subprocess fake-`xcrun` test now expects the anchored predicate string. A second real-subprocess test uses `--app-name MyAppDev --app-executable MyApp` (deliberately distinct) against a shim that only accepts the exact anchored argv, `processImagePath ENDSWITH "/MyAppDev.app/MyApp"`, then asserts on a parsed device record (mirroring the sibling test) and the absence of an `unexpected argv` stderr line -- closing the gap where a same-named fixture couldn't catch a bug that dropped `--app-executable` on its way to `startIosLogStream`. I confirmed this by ablating that line locally (removed `executableName: appExecutable` from the `startIosLogStream` call in `collector/run.ts`): the test timed out waiting for the parsed record; restoring the line made it pass again.
- Real-tool check (invariant 9): built this branch's CLI, then from `trailhead-bench` ran `worktree create qa-264 --carry-ignored`, and from the new worktree `start --json` and `ios --json` (fingerprint cache hit). After the launch settled, every record in `device.ndjson` carried `proc: "Trailhead"` (642/642 records with a `proc` field; the other two are the collector's own `collector_started`/`collector_stderr` events). `ps -ef | grep 'log stream'` showed the actual spawned command:
  ```
  simctl spawn <udid> log stream --style ndjson --predicate processImagePath ENDSWITH "/Trailhead.app/Trailhead"
  ```
  Then `stop --json` and, from `trailhead-bench`, `worktree remove <path> --force`; no `stim-*` simulators left behind (`stim-issue-245-measure` belongs to another session and was left alone).

Fixes #264

### DIFF
--- a/packages/stim-cli/src/__tests__/collector-parsers.test.ts
+++ b/packages/stim-cli/src/__tests__/collector-parsers.test.ts
@@ -124,8 +124,8 @@ describe('ios: log stream ndjson', () => {
     expect(parseLogStreamLine('{"eventType":"logEvent","messageType":"Default","eventMessage":"   "}')).toBe(null);
   });
 
-  test('logStreamArgs is the exact argv, with the predicate quoted for the device', () => {
-    expect(logStreamArgs('U1', 'MyApp')).toEqual([
+  test('logStreamArgs is the exact argv, anchored to the app executable path', () => {
+    expect(logStreamArgs('U1', 'MyApp', 'MyApp')).toEqual([
       'simctl',
       'spawn',
       'U1',
@@ -134,8 +134,32 @@ describe('ios: log stream ndjson', () => {
       '--style',
       'ndjson',
       '--predicate',
-      'processImagePath CONTAINS[c] "MyApp"',
+      'processImagePath ENDSWITH "/MyApp.app/MyApp"',
     ]);
+  });
+
+  test('logStreamArgs anchors to CFBundleExecutable when it differs from the .app basename', () => {
+    const args = logStreamArgs('U1', 'MyAppDev', 'MyApp');
+    expect(args[args.length - 1]).toBe('processImagePath ENDSWITH "/MyAppDev.app/MyApp"');
+  });
+
+  test('logStreamArgs falls back to the .app basename as the executable when none is given', () => {
+    expect(logStreamArgs('U1', 'MyApp')).toEqual(logStreamArgs('U1', 'MyApp', null));
+    expect(logStreamArgs('U1', 'MyApp')[8]).toBe('processImagePath ENDSWITH "/MyApp.app/MyApp"');
+  });
+
+  test('a short app name no longer admits an unrelated system process such as SpringBoard', () => {
+    const predicate = logStreamArgs('U1', 'app')[8];
+    assert(predicate);
+    const suffix = predicate.match(/ENDSWITH "(.+)"$/)?.[1];
+    assert(suffix);
+    const appsOwnPath =
+      '/Users/x/Library/Developer/CoreSimulator/Devices/U/data/Containers/Bundle/Application/ABC/app.app/app';
+    const springBoardPath = '/System/Library/CoreServices/SpringBoard.app/SpringBoard';
+    const amsengagementdPath = '/usr/libexec/amsengagementd';
+    expect(appsOwnPath.endsWith(suffix)).toBe(true);
+    expect(springBoardPath.endsWith(suffix)).toBe(false);
+    expect(amsengagementdPath.endsWith(suffix)).toBe(false);
   });
 
   test('appNameFromBundleId is the last segment, as a fallback for a caller with only a bundle id', () => {
--- a/packages/stim-cli/src/__tests__/collector-run.test.ts
+++ b/packages/stim-cli/src/__tests__/collector-run.test.ts
@@ -326,6 +326,48 @@ describe('the ios collector, spawned for real against a fake xcrun', () => {
     expect(deviceLog().some((r) => r.event === 'collector_stopped')).toBeTruthy();
   });
 
+  test('a distinct --app-executable anchors the predicate to it, not to --app-name', async () => {
+    // #264: the shim only accepts the anchor built from --app-executable, not --app-name.
+    writeShim(
+      'xcrun',
+      [
+        `case "$*" in`,
+        `  'simctl spawn UDID-1 log stream --style ndjson --predicate processImagePath ENDSWITH "/MyAppDev.app/MyApp"') ;;`,
+        `  *) echo "unexpected argv: $*" >&2; exit 9 ;;`,
+        `esac`,
+        ...iosShimLines()
+          .slice(0, 1)
+          .map((l) => `cat <<'LINE'\n${l}\nLINE`),
+        'exec sleep 30',
+      ].join('\n'),
+    );
+
+    const child = spawnCollector([
+      '--platform',
+      'ios',
+      '--root',
+      root,
+      '--udid',
+      'UDID-1',
+      '--bundle',
+      'com.example.app',
+      '--app-name',
+      'MyAppDev',
+      '--app-executable',
+      'MyApp',
+    ]);
+
+    const registered = await until(() => readCollectors(root).ios as CollectorEntry, {
+      label: 'the collector registration',
+    });
+    expect(registered.pid).toBe(child.pid);
+    expect(deviceLog().some((r) => r.event === 'collector_failed')).toBe(false);
+
+    process.kill(childPid(child), 'SIGTERM');
+    const result = await exited(child);
+    expect(result).toEqual({ code: 0, signal: null });
+  });
+
   test('survives the stream ending: a record, and exit 0', async () => {
     writeShim(
       'xcrun',
--- a/packages/stim-cli/src/__tests__/collector-run.test.ts
+++ b/packages/stim-cli/src/__tests__/collector-run.test.ts
@@ -106,6 +106,7 @@ describe('parseArgs', () => {
         udid: 'U1',
         bundleId: 'com.example.MyApp',
         appName: 'MyApp',
+        appExecutable: null,
         serial: null,
         packageName: null,
         physical: false,
@@ -135,6 +136,7 @@ describe('parseArgs', () => {
       udid: 'U1',
       bundleId: 'com.example.MyApp',
       appName: 'MyApp',
+      appExecutable: null,
       serial: null,
       packageName: null,
       physical: true,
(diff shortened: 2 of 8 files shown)


janicduplessis:
## Review: fix/collector-predicate-anchor (fixes #264)

Reviewed fresh in an isolated worktree (`git worktree add --detach ... origin/fix/collector-predicate-anchor`), no code written. Summary: **looks correct and mergeable**; a couple of small, non-blocking notes below.

### Correctness
- `processImagePath ENDSWITH "/<AppName>.app/<CFBundleExecutable>"` is sound. The leading `/` anchors on a directory boundary, so a short name like `App` does not falsely match `.../MyApp.app/App` or admit `SpringBoard.app/SpringBoard` or `/usr/libexec/amsengagementd` (verified by hand: the suffix comparison requires the char before the app name to be `/`, which excludes those). It also correctly handles the case where the `.app` basename and `CFBundleExecutable` differ (e.g. product name `MyAppDev` vs. executable `MyApp`), which the old `CONTAINS[c]` predicate didn't need to distinguish.
- Dropping `[c]` (case-insensitive) is fine/intentional here — both `appName` and `executableName` come from the built app's own path and Info.plist, i.e. ground truth, so exact-case `ENDSWITH` is strictly more correct than before.
- Nested bundles (extensions, watch app, app clips) are correctly excluded by design, matching the issue's stated intent.
- Not a regression, but not addressed either: neither the old nor new predicate escapes `"` (or other NSPredicate-significant characters) in `appName`/`exe` before interpolating into the quoted literal. Still low severity (app/product names rarely contain `"`), but worth a follow-up issue since this PR touches the exact line and doubles the interpolation surface (two values now, not one).

### `readBundleExecutable`/`parseBundleExecutable` (engine/xcode.ts)
Structurally and behaviorally identical to `readBundleId`/`parseBundleId` (plutil-JSON first, `defaults read` fallback, `null` on total failure, trims/type-checks the plist value). No divergence bugs found.

### Plumbing
Traced `commands/ios.ts` → `replaceCollector` → collector subprocess (`--app-executable`) → `collector/run.ts` `parseArgs` → `startIosLogStream` → `logStreamArgs`. Nothing dropped. `appExecutable` is computed once in `finishIosRun` and only forwarded on the simulator (`!physical`) `replaceCollector` call; the physical-device call (devicectl console, no `log stream`) correctly never receives it — that path doesn't need the predicate at all, consistent with invariant 2/3 discussion in the issue. Minor: `readBundleExecutable(appPath)` currently runs unconditionally whenever `appPath` is set, including on the physical-device path where the result is discarded — a small wasted `plutil`/`defaults` subprocess call on every device launch. Not a correctness issue, just avoidable work; could gate on `!physical`.

### Tests
Good, non-tautological coverage:
- `collector-parsers.test.ts` derives the predicate from the real `logStreamArgs` implementation and asserts against real system paths (`SpringBoard.app/SpringBoard`, `/usr/libexec/amsengagementd`) via the same `ENDSWITH`-suffix logic NSPredicate uses — this actually exercises the fix, not just structure.
- `engine-xcode.test.ts` mirrors the existing `readBundleId` test shape for the new function; exercises the plutil path, the `defaults` fallback, and total failure.
- `collector-run.test.ts` covers `--app-executable` parsing and updates the real-subprocess fake-`xcrun` test to the new predicate string end-to-end.
- `ios-command.test.ts` covers the plumbing from `finishIosRun` into `replaceCollector`, including the "CFBundleExecutable missing -> falls back to null" case.
- Gap (non-blocking): no test asserts that the **physical**-device `replaceCollector` call omits `appExecutable`. The existing physical tests (`the collector is the launch: it carries --physical...`) check `physical`/`udid` but not the absence of `appExecutable`. Worth locking in given that's the exact invariant the "physical path doesn't need this" claim rests on.

### Overlap with #261
The PR body's overlap note checks out. I fetched `fix/run-output-vocabulary` (#261) and ran `git merge-tree` between the two branches' merge-base — clean merge, no conflict markers, despite both touching `commands/ios.ts`. The two diffs don't touch the same lines (261 touches `phase(...)`/label text around installs and a new `deviceShortName` helper; 266 touches the `ReplaceCollectorArgs`/`IosDeps` shapes, `replaceCollector` body, and the `appName`/`appExecutable` locals in `finishIosRun`). One very minor wording nit: the PR body says the new local is "plumbed into the two existing calls," but the diff only forwards `appExecutable` on the simulator call site — the physical call site isn't touched at all (correctly, since it doesn't need it). Just imprecise phrasing, not a code issue.

### Real-tool check (invariant 9)
The PR body's transcript is specific and credible, not vague: named worktree (`qa-264`), exact commands (`worktree create --carry-ignored`, `start --json`, `ios --json`), a concrete result (642/642 `device.ndjson` records carrying `proc: "Trailhead"`, the 2 non-`proc` records identified as the collector's own `collector_started`/`collector_stderr` events), and the actual spawned argv confirmed via `ps -ef`:
```
simctl spawn <udid> log stream --style ndjson --predicate processImagePath ENDSWITH "/Trailhead.app/Trailhead"
```
That matches what `logStreamArgs` produces. I did not re-run this myself per the review scope, but the transcript reads as a real run, not fabricated.

### Checks run (fresh worktree, `pnpm install` + full suite)
- `pnpm run format:check` — pass
- `pnpm run lint` — pass
- `pnpm run build` — pass
- `pnpm run typecheck` — pass
- `pnpm run knip` — pass (one pre-existing, unrelated config hint: `pod` in `knip.json` `ignoreBinaries`)
- `pnpm test` — 3281/3283 passed on the full run; the 2 failures (`worktree.test.ts` "...exceeds 1MB" and `engine-xcode.test.ts` "discovers the workspace and resolves its scheme through a real -list -json") are both `Test timed out in 5000ms` under full-suite parallel load, and both pass individually/in isolation, including on `main` at HEAD. Confirmed unrelated to this diff — neither touched file nor either failing test is in the PR's changed files. Pre-existing flakiness, not introduced by this branch.

### Comment policy / invariants
No new comments in the diff at all — clean. No invariant violations found (single exec wrapper respected via `runFile`/executor pattern reused from `readBundleId`; no new commands/flags beyond the internal `--app-executable` collector-subprocess flag, which isn't user-facing).

**Verdict:** no blocking issues. The two "gap" notes (physical-path test coverage, unconditional `readBundleExecutable` call) are nice-to-haves, not required before merge.

janicduplessis:
Fix round: added a real-subprocess collector-run test with deliberately distinct --app-name/--app-executable (a same-named fixture couldn't have caught a bug that dropped --app-executable on its way to startIosLogStream), and a dim stderr note when CFBundleExecutable can't be read so a device log that falls back to the .app basename is diagnosable. Pushed as 4170b940e1a0507fee057360c1b94749da9b267c.

janicduplessis:
Second fix round: the distinct --app-executable test was a false positive (reject-branch exit 9 was recorded as a normal exit 0, and the accept branch's single fixture line never parsed into a device record, so nothing distinguished the branches). Fixed to cat the full fixture and assert on a parsed locationd record plus the absence of an 'unexpected argv' stderr line, mirroring the sibling test. Verified by ablating `executableName: appExecutable` at collector/run.ts:308 locally — the test timed out as expected — then restored it (clean diff after restore). Rebased onto origin/main (now past #262, with #261 merged). Rechecked the full battery from repo root under heavy local load (format:check, lint, build, typecheck, pnpm test x2 -- one run hit two load-induced timeouts in unrelated worktree-create.test.ts/android-command.test.ts, both untouched by this diff and clean on rerun, knip). Pushed as bb656b448d4c6aadf1a7024c838f9dd16870a3d9.

PR #360, opened 2026-09-04
fix: recover live bundle errors without relaunch

### LINKED ISSUE #351: Show Expo stack traces in logs --errors and guide JS reload recovery
A launch benchmark exposed two agent-loop gaps:

- Expo prints a console error header as an error record, then prints its symbolicated stack frames as separate informational Metro lines. stim logs --errors currently keeps only the header, so the useful stack is hidden.
- After fixing the JavaScript launch error, the agent ran stim ios a second time even though no native input changed. The shipped guidance should direct agents to Fast Refresh or a targeted Metro reload and reserve another platform command for native changes.

Keep one reported error per failure, preserve raw NDJSON in JSON mode, and add contract tests for the human stack output and reload guidance.

janicduplessis:
## Description

A JavaScript bundle-resolution failure leaves the React Native process alive but can prevent that process from handling Metro's reload command. The existing remedy therefore told an agent to run a command that appeared to succeed while leaving the error overlay unchanged, increasing the chance of an unnecessary second native launch.

## Solution

Classify the recovery by platform while preserving the existing no-rebuild boundary. Android remedies invoke React Native's exported native reload receiver on the exact Stim-owned emulator. iOS remedies inspect the exact Stim-owned simulator with agent-device and press the error overlay's Reload control. Ordinary runtime errors still try Metro reload first and provide the native path only as a fallback.

The agent guide carries the same distinction so bundle-build failures are not treated like native process crashes.

## Test plan

- `pnpm run format:check`
- `pnpm run lint`
- `pnpm run build`
- `pnpm run typecheck`
- `pnpm run knip`
- `pnpm test` — 92 files, 3,515 tests
- `pnpm run test:e2e` — 20 tests
- `pnpm run test:runtime`
- Bare React Native 0.85.3 on iOS: introduced an unresolved startup import, confirmed Stim reported the live-process bundle error with stack/code frame, repaired it, then recovered the unchanged starter screen by pressing the overlay Reload control through agent-device on the same simulator without a second `stim ios`.
- Bare React Native 0.85.3 on Android: introduced the same unresolved startup import, repaired it, then recovered the unchanged starter screen with `<package>.RELOAD_APP_ACTION` on the same emulator without a second `stim android`.

Fixes #351

### DIFF
--- a/packages/stim-cli/src/__tests__/android-command.test.ts
+++ b/packages/stim-cli/src/__tests__/android-command.test.ts
@@ -2481,6 +2481,8 @@ describe('launch verification', () => {
     expect(text).toMatch(/^  launch {6}a redbox from the app$/m);
     expect(text).not.toMatch(/a native framework error/);
     expect(text).toContain('agent-device metro reload --metro-port 8082');
+    expect(text).toContain('adb -s emulator-5584 shell am broadcast -a com.example.app.RELOAD_APP_ACTION');
+    expect(text).toMatch(/invoke React Native's native reload receiver/);
     expect(text).toMatch(/Do not run `stim android` unless native inputs changed or the app process exits/);
   });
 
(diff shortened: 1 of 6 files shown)


janicduplessis:
Closing in favor of implementing a first-class stim reload command. That command will own platform-specific recovery while preserving the no-build/no-native-launch contract, so these agent-facing workaround instructions would be redundant.

PR #405, opened 2026-09-05
perf: route Android CMake compiles through ccache

### LINKED ISSUE #398: Route Android CMake compiles through ccache with path mapping so C++ objects cross worktrees
## Problem

`stim android` never reuses a C++ object across worktrees. Every AGP CMake task (`configureCMakeDebug[abi]`, `buildCMakeDebug[abi]`) is `@DisableCachingByDefault`, so `--build-cache` cannot serve it, and Stim sets no compiler launcher. On trailhead (Expo SDK 57, RN 0.86, new architecture, nine CMake modules) C++ is 89% of summed task time in a fresh worktree's first build, and the 387 tasks the Gradle build cache does serve total 9 s. iOS has the Xcode compilation cache under `~/.stim/compilation-cache`; Android has no counterpart.

React Native's own `ReactNative-application.cmake` already routes the `:app` CMake project through ccache when it is on PATH, but with ccache's defaults (`hash_dir=true`, no `base_dir`) the working directory and every `-I<worktree>/node_modules/...` path enter the hash, so a second worktree gets zero hits, and the eight library CMake projects have no launcher at all.

## Evidence

Measured on trailhead at 8ae44bd with stim rc.13 (logs, profiles, ccache call lists and APKs under the session scratchpad `android-profile/`, 26 builds):

- Fresh worktree, four ABIs, no launcher (build A): 215 s wall; C++ task time 600 s of 673 s.
- Same ABI (arm64), second worktree, `.cxx` deleted, launcher via environment with `CCACHE_BASEDIR` and default `hash_dir`: 0 of 380 hits. ccache debug replay: the two hash inputs differ only in the `### cwd` line.
- Same with `CCACHE_NOHASHDIR=true`: 176 of 380 hits; wall 47.6 s -> 30.2 s; `buildCMakeDebug[arm64-v8a]` of the six hitting modules 54.6 s -> 2.1 s (app, skia, nitro-modules, mmkv, screens, gesture-handler at 100%).
- Fresh worktree (`git worktree add` + `npm ci`), arm64 only, primed cache, load 4: 34.2 s wall, versus 49.6 s for the same without the launcher; with every object hitting and the `createExpoConfig` stall removed, 11.2 s.
- The 204 misses are exactly the three modules that use `target_precompile_headers`: reanimated (106 TUs), expo-modules-core (60), worklets (38). CMake writes an absolute `#include` into `cmake_pch.hxx` and the `.pch` embeds 407 worktree path strings, so their hash can never match across paths. Separate follow-up.
- Producer-side cost is nil (47.6 s with launcher versus 40 to 61 s without, same state). Cache size: 212 MB per worktree-equivalent of arm64 objects, 350 MB with the PCH duplicates.
- Mechanism: the NDK r27/r28 CMake toolchain reads `NDK_CCACHE` / `ANDROID_CCACHE` only as CMake variables (no environment route), but CMake 3.22 seeds `CMAKE_<LANG>_COMPILER_LAUNCHER` from the environment variables of the same name on a fresh configure, and AGP forwards Gradle's environment to CMake. All nine `CMakeCache.txt` files carried the launcher after one configure. `ccache ccache clang` (RN's own `RULE_LAUNCH_COMPILE` plus the launcher) nests without error.

## Cause

`engine/gradle.ts` spawns `gradlew` with the inherited environment plus `TERM` and `FORCE_COLOR`; nothing selects a compiler launcher or a shared, relocatable ccache configuration.

## Fix idea

When a `ccache` binary is on PATH, set on the Gradle child environment in `buildAndroid`:

- `CMAKE_C_COMPILER_LAUNCHER` and `CMAKE_CXX_COMPILER_LAUNCHER` = the absolute ccache path (never a bare name; agent shells can lack `/opt/homebrew/bin`).
- `CCACHE_DIR` = `$STIM_HOME/ccache` (new helper in `paths.ts` next to the compilation cache; the unused `sharedGradle()` is the natural spot), registered through `cache-manifest.ts` like the CAS so `gc` reports it.
- `CCACHE_BASEDIR` = the workspace root (realpath). Paths under it are rewritten relative to the compile cwd, which has the same shape in every worktree; paths outside it (NDK sysroot, `~/.gradle` prefab includes) are identical across worktrees anyway.
- `CCACHE_NOHASHDIR=true` (exact spelling; `CCACHE_HASHDIR=false` is not honored). Trade-off: an object reused from worktree A carries A's directory as DWARF `comp_dir` in B; same class as the CAS trade-off doctor already explains for iOS.
- `CCACHE_SLOPPINESS`: at least `pch_defines,time_macros`; whether `include_file_mtime,include_file_ctime` are needed for the hits above, and whether a same-path PCH hit after a header mtime change breaks worklets and expo-modules-core (they lack `-fno-pch-timestamp`; reanimated passes it), is being measured and will be posted as a comment here. Pick the value from that comment.
- `CCACHE_MAXSIZE` explicit (ccache prunes itself at that limit; gc reports the size).

Print one dim `cache  ccache on (<dir>)` line like `xcode.ts` does for the CAS, and report per-build hits and misses in the `cache` block and the `--json` facts, mirroring `compilationCacheActivity` (ccache's `stats_log` / `CCACHE_STATSLOG` gives a per-build count without zeroing shared stats).

Preserve: no Gradle argument changes, no init script, no cache-key change (outputs are byte-identical, so invariant 10 and `buildCacheKey` are untouched). Do not touch ABI selection (PR #397 owns it). The launcher persists in each workspace's `.cxx/**/CMakeCache.txt` after the first configure, so non-Stim `./gradlew` runs in that checkout also go through ccache, and a `.cxx` that predates the variables keeps compiling without it: say both in the guide, and add doctor findings for ccache absent (`brew install ccache`), for a `.cxx` whose `CMakeCache.txt` names a launcher that is not on disk, and qualify the iOS finding that says ccache "keys on absolute paths, so it misses across worktrees anyway" (true only with the default configuration). Verify for real (invariant 9): two worktrees of a project with CMake modules, second build shows hits in `ccache -s` and a shorter `buildCMakeDebug`.

janicduplessis:
## Description

`stim android` never reused a C++ object across worktrees. Every AGP CMake task
(`configureCMakeDebug[abi]`, `buildCMakeDebug[abi]`) is uncacheable by Gradle, so
`--build-cache` serves none of them, and `engine/gradle.ts` spawned `gradlew` with
the inherited environment plus `TERM` and `FORCE_COLOR` -- nothing selected a
compiler launcher. On a React Native app with native modules that leaves a fresh
worktree recompiling every C++ translation unit, which is the bulk of its first
Android build. iOS has had the Xcode compilation cache under
`~/.stim/compilation-cache` since Xcode 26; Android had no counterpart.

React Native's own `ReactNative-application.cmake` already routes the `:app` CMake
project through ccache when it is on PATH, but with ccache's defaults
(`hash_dir=true`, no `base_dir`) the working directory and every absolute include
path enter the hash, so a second worktree gets no hits -- and the library CMake
projects get no launcher at all.

## Solution

When a `ccache` binary resolves on PATH, `buildAndroid` spawns Gradle with a
computed environment, and **nothing else changes**: same argument list, same cache
key, no Gradle init script, no new command or flag.

The new `engine/ccache.ts` keeps the pure part separate. `ccacheEnvironment()`
returns the eight variables (they are printed verbatim in the evidence below);
`resolveCcache()` is the thin I/O wrapper that resolves the binary through the
single exec wrapper, declines to fight a project that passes its own
`CMAKE_*_COMPILER_LAUNCHER` in `android/app/build.gradle`, registers
`$STIM_HOME/ccache` in the cache manifest so `gc` reports it, and prints one dim
`cache  ccache on (<dir>)` line the way `xcode.ts` does for the CAS. After the
build the per-workspace `CCACHE_STATSLOG` is parsed into hits and misses -- a
per-build count that does not zero the shared statistics -- and reported as the
existing `compilation cache` label in the run summary and as `ccache` in the
`--json` facts, so the closed label vocabulary in `command-output.ts` is unchanged.

**Two costs, both deliberate and both documented in `guide lifecycle`.**
`CCACHE_NOHASHDIR` is what lets an object cross paths at all, and it means an
object reused from worktree A carries A's directory as its DWARF `comp_dir` --
the same class of trade-off doctor already explains for the iOS CAS. And the
modules that precompile a header without `-fno-pch-timestamp` -- worklets and
expo-modules-core -- recompile whenever the mtime of a PCH input changes, which
a fresh `npm ci` in a new worktree does: ccache rebuilds the header and every
translation unit including it misses, 5 to 11 seconds per module. No stale
`.pch` is served. reanimated passes `-Xclang -fno-pch-timestamp` and hits
across worktrees.

`CCACHE_SLOPPINESS` is `pch_defines,time_macros`, the value the issue's
"Verification results" comment measured: dropping `include_file_mtime,
include_file_ctime` lost no hits (380 of 380 direct hits against entries stored
under the wider setting), and PCH reuse under it was confirmed safe.

The launcher persists in the checkout: AGP writes it into every
`.cxx/**/CMakeCache.txt` on the first configure, so a plain `./gradlew` there also
compiles through ccache, and a `.cxx` written before the variables existed keeps
compiling without them. `doctor` reports both, as Android findings that honor
`--platform`: a "costs time" finding when ccache is not on PATH (with the measured
cost and the delete-`.cxx` remedy, per the maintainer requirement on the issue),
and one when a `CMakeCache.txt` names an absolute launcher that is not on this
machine, or names none at all (a bare name is CMake's own PATH lookup, so it is
left alone). The iOS `checkCcacheConflict` detail no longer claims ccache misses across
worktrees unconditionally -- that is true of its default configuration.

## Test plan

**Result first:** a second worktree of a real app built in **35s against the
producer's 55s**, with **176 of 380 C++ compiles served from the other worktree's
objects**.

### Real tool verification (invariant 9)

Two throwaway `git worktree` checkouts of an Expo SDK 57 / RN 0.86 app with nine
CMake modules (trailhead at `b0e4f23`), `npm ci` in each, a temporary `STIM_HOME`
so `~/.stim` was untouched, and the environment read out of the built
`resolveCcache()` rather than retyped. Same command in both:

```
./gradlew assembleDebug --build-cache --console=plain -PreactNativeArchitectures=arm64-v8a
```

**Producer** (`ccache-a`, empty cache). What Stim set:

```
CMAKE_C_COMPILER_LAUNCHER=/opt/homebrew/bin/ccache
CMAKE_CXX_COMPILER_LAUNCHER=/opt/homebrew/bin/ccache
CCACHE_DIR=<temp STIM_HOME>/ccache
CCACHE_BASEDIR=/Users/janicduplessis/Developer/trailhead-worktrees/ccache-a
CCACHE_NOHASHDIR=true
CCACHE_SLOPPINESS=pch_defines,time_macros
CCACHE_MAXSIZE=5G
CCACHE_STATSLOG=<temp STIM_HOME>/workspaces/ccache-a--68830000f168da8d/logs/ccache-stats.log
```

`BUILD SUCCESSFUL in 55s`. `ccache -s`: 380 cacheable calls, 10 hits / 370 misses,
0.2 GB. The stats-log parser read the same numbers: `10 hits / 370 misses (2.6%)`.
One configure was enough to persist the launcher into all seven library modules --
e.g. `node_modules/react-native-reanimated/android/.cxx/Debug/b4a3sy2o/arm64-v8a/CMakeCache.txt`
carries `CMAKE_CXX_COMPILER_LAUNCHER:STRING=/opt/homebrew/bin/ccache`.

**Consumer** (`ccache-b`, a different absolute path, its own `.cxx`, same warm
cache; `--profile` added to this run only, to read per-task times). 35s wall.
`ccache -s` went 380 -> 760 cacheable calls and 10 -> 186 hits, i.e. 176 hits and
204 misses for this build -- which is exactly what the parser reported and what
`--json` carried:

```
{"status":"reported","hits":176,"misses":204,"hitRatePercent":46.3}
```

Per-task, the six modules that hit collapse to a second or so; the three
`target_precompile_headers` modules are the misses:

```
:react-native-gesture-handler:buildCMakeDebug[arm64-v8a]         0.442s
:react-native-screens:buildCMakeDebug[arm64-v8a]                 0.467s
:react-native-nitro-modules:buildCMakeDebug[arm64-v8a]           0.674s
:react-native-mmkv:buildCMakeDebug[arm64-v8a]                    0.721s
:shopify_react-native-skia:buildCMakeDebug[arm64-v8a]            1.190s
:app:buildCMakeDebug[arm64-v8a]                                  1.348s
:react-native-worklets:buildCMakeDebug[arm64-v8a][worklets]      5.256s
:expo-modules-core:buildCMakeDebug[arm64-v8a]                   10.637s
:react-native-reanimated:buildCMakeDebug[arm64-v8a][reanimated] 17.690s
```

That 176/204 split matches the 176-of-380 the issue predicted, and the misses in
this run are the three `target_precompile_headers` modules (each worktree got its
own `npm ci`, so every PCH input had a new mtime).

**End to end with the built CLI** in `ccache-b` (temporary `STIM_HOME`, owned
emulator created, then shut down and deleted):

```
$ stim start && stim android --json
  build       compiling debug with Gradle
  cache       ccache on (<temp STIM_HOME>/ccache)
  cache       gradle build cache on (--build-cache, shared under the Gradle user home)
  build       ok (17.6s)
  launch      com.appandflow.trailhead (375ms)
...,"ccache":{"status":"reported","hits":0,"misses":2,"hitRatePercent":0},...

$ stim android                 # second run, artifact cache hit
  cache       cache hit
  compilation cache not run; artifact cache supplied the app

$ stim gc
        350M  ccache (registered)
              ccache keeps itself under CCACHE_MAXSIZE, so it is emptied whole or not at all
```

`stim logs --errors` was clean and `stim stop` shut the emulator down. 350 MB for
one arm64 Debug worktree pair matches the size the issue measured, PCH duplicates
included. Both trailhead worktrees were removed afterwards.

`doctor` on a real project, with and without `/opt/homebrew/bin` on PATH:

```
$ PATH=/usr/bin:/bin:/usr/sbin:/sbin stim doctor --platform android
costs time  ccache is not on PATH, so Android C++ recompiles in every worktree
  -> brew install ccache, then delete android/app/.cxx and
     node_modules/**/android/.cxx once so CMake reconfigures with the launcher. ...

$ stim doctor --platform android          # ccache on PATH, and in ccache-a after the build
(no ccache finding)
```

### Suites

`pnpm run format:check`, `lint`, `build`, `typecheck`, `knip` clean;
`NODE_OPTIONS=--use-openssl-ca pnpm test` 3642 tests across 98 files;
`pnpm run test:e2e` 24/24. New coverage in `engine-ccache.test.ts` (exact variable
set, absolute-path locator, project-launcher gate, and the stats-log parser
against a fixture captured from real ccache 4.12.3 output), plus the environment
reaching the Gradle child in `engine-gradle.test.ts` -- including that without
ccache the argv and environment are byte-identical to before -- and the facts,
doctor and guide contracts.

Fixes #398

### DIFF
--- a/packages/stim-cli/src/commands/android.ts
+++ b/packages/stim-cli/src/commands/android.ts
@@ -12,7 +12,7 @@ import {
 } from '@stim-cli/cache';
 import type { FingerprintSource } from '@expo/fingerprint';
 import { formatDuration, phaseLine, shortHash, SLOW_STEP_MS, stepClock, stepTimer } from '../command-output.ts';
-import type { RemoteDeviceBackend, WaitedForBuild } from '../types.ts';
+import type { CcacheActivity, RemoteDeviceBackend, WaitedForBuild } from '../types.ts';
 import {
   appProjectProblem,
   findProjectRoot,
@@ -119,6 +119,7 @@ import { detectProviders } from '../engine/metro-reach.ts';
 import { selectFromPool } from '../engine/device-pool.ts';
 import { needsPrebuild, runPrebuild } from '../engine/prebuild.ts';
 import { buildAndroid, productFlavorRefusal, readProductFlavors } from '../engine/gradle.ts';
+import { CCACHE_NOT_RUN, CCACHE_UNAVAILABLE, resolveCcache } from '../engine/ccache.ts';
 import { swapApkBundle, resolveKeystore } from '../engine/apk-swap.ts';
 import { captureAssetManifest } from '../engine/asset-manifest.ts';
 import {
@@ -338,6 +339,7 @@ interface RunAndroidOptions {
   needsPrebuildFor?: typeof needsPrebuild;
   prebuild?: typeof runPrebuild;
   build?: typeof buildAndroid;
+  ccacheFor?: typeof resolveCcache;
   install?: typeof installAndroidApp;
   launch?: typeof launchAndroidApp;
   launchRelease?: typeof launchAndroidReleaseApp;
@@ -416,6 +418,7 @@ function resolveRunAndroidOptions(
     needsPrebuildFor = needsPrebuild,
     prebuild = runPrebuild,
     build = buildAndroid,
+    ccacheFor = resolveCcache,
     install = installAndroidApp,
     launch = launchAndroidApp,
     launchRelease = launchAndroidReleaseApp,
@@ -493,6 +496,7 @@ function resolveRunAndroidOptions(
     needsPrebuildFor,
     prebuild,
     build,
+    ccacheFor,
     install,
     launch,
     launchRelease,
@@ -572,6 +576,7 @@ export async function runAndroid(options: RunAndroidOptions = {} as RunAndroidOp
     needsPrebuildFor,
     prebuild,
     build,
+    ccacheFor,
     install,
     launch,
     launchRelease,
@@ -1039,6 +1044,7 @@ export async function runAndroid(options: RunAndroidOptions = {} as RunAndroidOp
   let storeKey = '';
   let storeSources: FingerprintSource[] = [];
   let apkPath: string | null = null;
+  let ccacheActivity: CcacheActivity = CCACHE_NOT_RUN;
 
   async function resolveInitialFingerprint(): Promise<boolean> {
     const fingerprintTimer = stepTimer(now);
@@ -1373,8 +1379,9 @@ export async function runAndroid(options: RunAndroidOptions = {} as RunAndroidOp
             phase('build', `compiling ${variant || 'debug'} with Gradle`);
             const built: BuildAndroidResultLike = await build(
               { root, logWriter: writer, variant, abi: buildAbi },
-              { estimateMs: estimates().coldBuildMs },
+              { estimateMs: estimates().coldBuildMs, ccache: ccacheFor({ root, onNote: out }) },
             );
+            ccacheActivity = built.ccache ?? CCACHE_UNAVAILABLE;
             if (built.failed) {
               const diagnostics = built.diagnostics || [];
               for (const diag of diagnostics) {
@@ -1548,6 +1555,7 @@ export async function runAndroid(options: RunAndroidOptions = {} as RunAndroidOp
         swapDir,
         record,
         waitedForBuild,
+        ccache: ccacheActivity,
         uploadPending,
         providerUpload,
         providerName,
--- a/packages/stim-cli/src/__tests__/android-command.test.ts
+++ b/packages/stim-cli/src/__tests__/android-command.test.ts
@@ -342,8 +342,15 @@ function harness(overrides = {}) {
     build: async (args: BuildArgs = {}) => {
       calls.order.push('build');
       calls.build.push(args);
-      return { ok: true, apkPath: fakeApk(), durationMs: 161000, lastLines: [] };
+      return {
+        ok: true,
+        apkPath: fakeApk(),
+        durationMs: 161000,
+        lastLines: [],
+        ccache: { status: 'reported' as const, hits: 176, misses: 204, hitRatePercent: 46.3 },
+      };
     },
+    ccacheFor: () => null,
     install: (args: InstallArgs = {}) => {
       calls.install.push(args);
       return { ok: true, apkPath: args.apkPath ?? '' };
@@ -1030,6 +1037,7 @@ describe('a cache hit', () => {
       bundleId: 'com.example.app',
       installSkipped: false,
       launched: true,
+      ccache: { status: 'not-run', hits: null, misses: null, hitRatePercent: null },
       debugHttpHost: '10.0.2.2:8082',
       debugHttpHostNote: null,
       devClientUrl: null,
@@ -1057,6 +1065,34 @@ describe('a cache miss', () => {
     expect(labelled(h.stderr, 'build')[1]).toMatch(/^  build {7}ok \(2m41s\)$/);
     assert(result.facts);
     expect(result.facts.cacheHit).toBe(false);
+  });
+
+  test('hands the resolved ccache environment to Gradle and reports its counts', async () => {
+    const setup = {
+      dir: join(home, 'ccache'),
+      statsLog: join(home, 'ccache-stats.log'),
+      env: { CMAKE_CXX_COMPILER_LAUNCHER: '/opt/homebrew/bin/ccache' },
+    };
+    const options: Record<string, unknown>[] = [];
+    const h = harness({
+      ccacheFor: () => setup,
+      build: async (_args: BuildArgs = {}, opts: Record<string, unknown> = {}) => {
+        options.push(opts);
+        return {
+          ok: true,
+          apkPath: fakeApk(),
+          durationMs: 161000,
+          lastLines: [],
+          ccache: { status: 'reported' as const, hits: 176, misses: 204, hitRatePercent: 46.3 },
+        };
+      },
+    });
+    const result = await h.run();
+    expect(result.ok).toBe(true);
+    expect(options[0]?.ccache).toBe(setup);
+    assert(result.facts);
+    expect(result.facts.ccache).toEqual({ status: 'reported', hits: 176, misses: 204, hitRatePercent: 46.3 });
+    expect(h.stdout.join('\n')).toMatch(/^ {2}compilation cache 176 hits \/ 204 misses \(46\.3%\)$/m);
   });
 
   test('builds only the ABI selected for an owned emulator and scopes the cache key', async () => {
@@ -2461,6 +2497,7 @@ describe('the pure parts', () => {
       bundleId: null,
       installSkipped: false,
       launched: false,
+      ccache: { status: 'unavailable', hits: null, misses: null, hitRatePercent: null },
       debugHttpHost: null,
       debugHttpHostNote: null,
       devClientUrl: null,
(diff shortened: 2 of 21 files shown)


packages/stim-cli/src/commands/android.ts
```diff
@@ -2595,8 +2612,9 @@ export async function runAndroid(options: RunAndroidOptions = {} as RunAndroidOp
             phase('build', `compiling ${variant || 'debug'} with Gradle`);
             const built: BuildAndroidResultLike = await build(
               { root, logWriter: writer, variant, abi: buildAbi },
-              { estimateMs: estimates().coldBuildMs },
+              { estimateMs: estimates().coldBuildMs, ccache: ccacheFor({ root, onNote: out }) },
```
janicduplessis, line 1382: Resolution happens here rather than inside `buildAndroid`, which is where `xcode.ts` resolves the CAS. The reason is invariant 5: `resolveCcache` writes to the cache manifest and probes PATH, and every existing `buildAndroid` unit test calls the engine without a temporary `STIM_HOME`, so resolving inside the engine would have written to the real `~/.stim/caches.json` from the test suite and made the notes machine-dependent. Keeping the default `null` in the engine leaves those tests exercising the same argv and environment as before, and `ccacheFor` is injectable for the command tests.

janicduplessis:
CI note: the `pull_request`-triggered run never appeared for this PR (no check suite on the head commit 15+ minutes after it was opened, while other PRs in the repo got one immediately). I dispatched `ci.yml` manually against the branch instead -- https://github.com/appandflow/stim/actions/runs/33945115301 -- and all three jobs pass on 87ef685: `published runtime (node 20.19.4)`, `test (node 22)`, `test (node 24)`. Marking the PR ready for review should trigger the normal PR run.

PR #446, opened 2026-09-06
fix: reuse Android PCHs across worktrees

### LINKED ISSUE #444: Android PCH cache reuse retains worktree-specific paths
## Problem

Android C++ cache reuse remains poor across warmed worktrees even after stale CMake compiler-launcher state is repaired. This is distinct from #440 (missing launchers) and #441 (benchmark validation).

A clean diagnostic build recorded 112 hits / 205 misses (35.3%). The immediately preceding golden worktree build recorded 6 hits / 160 misses (3.6%). The cache is populated and the compiler calls do go through ccache. We should not attribute these measured misses to bypassing the launcher.

## Evidence

- Stim 1.0.0-rc.16; current origin/main d6a484f has the same `src/engine/ccache.ts` and `src/engine/gradle.ts` as the diagnostic build. Doctor-only changes from #442 prepared the fixture outside the measured build.
- Apple M4 / 16 GB, ccache 4.13.6, Gradle 9.3.1, Android NDK 27.1.12297006, ARM64 only.
- Trailhead fixture 63fe92cb06e228fb0294344a67778180392f3008; RN 0.86.3, Reanimated 4.5.1, Worklets 0.10.1.
- Repro: build the main checkout with ccache enabled; warm/build a separate worktree into the shared ccache; remove the warm-up worktree normally; warm/build another worktree against a copy of that populated cache with `CCACHE_DEBUG=true` and `CCACHE_DEBUGDIR` outside the checkout.
- Diagnostic build: 66.7 seconds, 868 Gradle tasks (220 executed / 185 from cache / 463 up-to-date). Cache miss breakdown: Reanimated 106, Expo Modules Core 60, Worklets 38, Gesture Handler 1. All other C++ calls hit.
- Expo PCH debug log retrieves a prior manifest, then reports `<removed-warmup-worktree>/node_modules/expo/node_modules/expo-modules-core/android/src/main/cpp/ExpoHeader.pch is mentioned in a manifest entry but can't be read (No such file or directory)`, followed by `Did not find result key in manifest` and `Not considering cached precompiled header in preprocessor mode`.
- Reanimated/Worklets compile from packages resolved under the main checkout. Their preprocessed command still contains a relative include path with the diagnostic worktree name, e.g. `-I../../../../../../../../trailhead-v4-worktrees/bench-ccache-diagnostic/node_modules/react-native/ReactCommon`; neither PCH direct-lookup key exists in the populated cache.
- Local evidence preserved under benchmark campaign `state/ccache-diagnostic/result.json` and `state/ccache-diagnostic/debug/`. No timed agent results were published from these probes.

## Cause

Confirmed for Expo: a cached PCH manifest depends on a removed worktree's absolute header path, preventing direct reuse. For Reanimated/Worklets, worktree identity survives path rewriting because the compile working directory is in the shared main-checkout dependency while some headers resolve through the new worktree. PCH rebuilding then coincides with most downstream misses. Further controlled input comparisons are needed before claiming one complete fix for all three packages.

`CCACHE_BASEDIR` and `CCACHE_NOHASHDIR` alone do not eliminate these path dependencies. Increasing warm-up repetitions or weakening a benchmark threshold is not a fix.

## Fix idea

Investigate relocatable compiler inputs/PCH dependencies and package realpath handling. Preserve cache correctness and isolated worktrees; do not ignore header content or timestamp checks indiscriminately. Verify by building in two differently named worktrees, removing the first, and showing healthy reuse in the second using per-invocation ccache debug evidence. Keep a benchmark investigation flag until a representative warm baseline is established.

janicduplessis:
## Description

Android PCHs retain checkout-specific paths even with timestamp fixes and ccache enabled. A cached PCH can also refer to an old AGP build-directory hash, making the next checkout's build fail. Fixes #444.

## Solution

Enabled automatically for eligible Android Clang 18 / CMake 3.22+ targets; an incompatibility could fail their native build. Custom compiler setups and unsupported targets keep their existing behavior.

A Gradle init script enables relocation without editing dependency sources. Generated PCH wrappers are staged by content and include exact headers through private macros, preserving search order and system-header warnings. Only workspace filenames in preprocessor linemarkers are normalized for ccache; source tokens and header-content validation stay intact. The wrapper identity participates in the cache key, and configured native targets continue working without Stim's environment.

PCH and [#405's timestamp handling](https://github.com/appandflow/stim/commit/b9743cd32a76dbd979dd1b54417a66757d30830e) stay enabled; the guide's timestamp-only explanation is corrected to cover the reproduced path dependency.

## Test plan

- Run `test/e2e/pch-relocation.e2e.js` with `STIM_PCH_TEST_NDK` and `STIM_PCH_TEST_CMAKE` set: verifies reuse after the original checkout disappears, header-edit invalidation, exact header/search/warning behavior, and preservation of custom, disabled, symlinked and mixed-language targets.
- `test/e2e/pch-compiler.e2e.js` checks preprocessing-file/stdout normalization without rewriting source tokens, external paths, failed output or unsupported encodings.
- SDK 58 Trailhead, NDK 27.1 / Clang 18, AGP 9.2.1, arm64: two fresh copies built APKs through `buildAndroid` with private caches and the first checkout made unavailable. The replay took 56 seconds with **389 hits / 0 misses**, including PCHs. This is a build diagnostic, not an agent benchmark; device installation and UI were not exercised.


janicduplessis:
Closing this relocation proof of concept: we are going with disabling Android PCH for now, in a separate PR. The simpler approach avoids maintaining the relocation machinery; we have not demonstrated a warm-cache speed benefit for relocation over disabling PCH. Keeping this PR and its branch as a reference if we revisit that tradeoff later.

janicduplessis:
The independent disable-PCH replacement is now #447. This relocation branch remains preserved as a proof of concept.

PR #466, opened 2026-09-07
fix: isolate benchmark runners from coordinator data

### LINKED ISSUE #460: Audit benchmark reads outside the allowed run scope
## Problem

The benchmark collector can accept a run whose agent reads coordinator files outside the prompt's allowed paths. Device isolation and Settings proof can pass despite that scope violation.

## Evidence

During the Android campaign tracked by #450, manual review found Opus native control ran `cat <campaign-root>/env.sh` and `cat <campaign-root>/pins.env`. The prompt permits the fixture, run worktree, current run evidence directory, and explicit proof scratch paths, not the coordinator root. The automatic audit accepted the run, but it is being excluded manually before publication. No credential contents were read in the observed command.

## Cause

The collector audits device sessions, setup commands, tool provenance, cache evidence, and proof. It does not enforce the prompt's coordinator-file read boundary.

## Fix idea

Add auditable detection for actual shell/file-tool reads of coordinator config, other runs, and golden state. Preserve necessary SDK/skill reads, quoted searches, and current-run proof paths; do not use an indiscriminate substring scan. Consider an explicit persisted manual-audit override so a later re-collection cannot erase a reviewed exclusion. Add regressions from the observed command and legitimate neighboring paths.

janicduplessis:
## Description

Prompt-only scope rules leave benchmark agents able to read coordinator configuration and sibling results. Enforce that boundary for both runners instead of discovering contamination after app proof succeeds.

## Solution

Launch Codex and Claude under the same macOS filesystem policy: protected file contents and writes are denied while preserving access to the current run, required tools and shared state. Parent-directory listings remain allowed. Configured tool/cache paths cannot overlap private benchmark directories. Real read/write, child-process and symlink probes run before dispatch; a failed check prevents timing from starting. Preserve the verified policy digest with each run.

This prevents accidental access through the runner process tree, not malicious access through already-running native services. It relies on macOS's deprecated `sandbox-exec` launcher and fails closed when unavailable. Historical audit verdicts and benchmark execution rules are unchanged.

## Test plan

- Real macOS tests deny coordinator/golden/sibling reads and writes, including symlink aliases and children; allow current-run writes, read-only tool access, and creation of the exact future Git worktree.
- Confirm broad or narrow private-directory cache grants, including symlink aliases, and modified verified policies refuse launch.
- Untimed Stim/control smoke checks pass with the installed Codex skill profile, and the installed Claude executable starts under both policies. No model benchmark was rerun.

Fixes #460

### DIFF
--- a/docs/agent-benchmark.md
+++ b/docs/agent-benchmark.md
@@ -100,6 +100,22 @@ independently pinned `agent-device` skill. The control profile exposes only
 `agent-device`; the Stim binary and skill are unavailable. Both profiles have
 the same model settings, filesystem authority, and app task.
 
+The macOS coordinator launches both runners and their child processes under
+the same filesystem policy. It denies coordinator configuration, golden-state
+and sibling-run file contents and writes, while permitting parent-directory
+listings. Each run retains its own worktree, proof, temporary files, runner
+home, selected tools, and the configured shared native-cache/device state.
+Coordinator metadata and transcripts remain outside those writable grants.
+Before timing starts, real read/write and child-process probes must verify the
+policy; an unavailable sandbox or overbroad grant refuses dispatch. The policy
+and its digest are retained with the private run metadata.
+
+This boundary prevents accidental benchmark-data access, not adversarial host
+access: native system services and pre-existing processes are not sandboxed by
+the runner policy. Strong isolation from a malicious agent requires a separate
+host or VM. The coordinator still independently verifies app/device evidence.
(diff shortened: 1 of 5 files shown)


janicduplessis:
Independent code review is clear after rejecting configured grants that overlap private directories, including narrow descendants and aliases. Verified 3,617 tests, 66 e2e tests, all repository checks, real macOS sandbox and Git worktree tests, plus untimed Stim/control runner-profile preflights. No model benchmark was rerun.

PR #468, opened 2026-09-07
docs: require warm completion before starting the app

### LINKED ISSUE #467: Clarify that worktree warming must finish before starting the app
## Problem

The agent guide orders warm before start but does not explicitly distinguish a shell tool returning a running-job ID from warm completing. Starting Metro or installing dependencies while ignored state is still being copied can race preparation.

## Evidence

On current main, packages/stim-cli/src/guide/agent.ts shows `stim worktree warm` followed by `stim start`, without an explicit completion requirement. The detailed lifecycle options describe copy failures and empty stdout but not waiting on a yielded shell job.

## Cause

The warm command awaits copying and reports completion on stderr. A shell harness can yield before that command exits; the guide leaves that distinction implicit.

## Fix idea

Require successful warm completion before start, platform builds, or dependency installation. Tell agents to poll/wait for a yielded job's exit status, and to resolve failed or incomplete warming before continuing. Keep operational instructions in the version-matched agent/lifecycle guides; leave the static skill router unchanged.

janicduplessis:
## Description

A shell tool can return a running-job ID while worktree warming is still copying dependencies. The guide's command order alone does not tell agents to wait before starting Metro or installing packages.

## Solution

Make successful warm completion an explicit prerequisite for start, platform builds, and dependency installation. Explain how to handle yielded jobs in the version-matched agent and lifecycle guides; keep the static skill router and runtime behavior unchanged.

## Test plan

- Render `stim guide agent` and `stim guide lifecycle options`; verify both require warm to exit successfully and distinguish a running job from completion.
- Extend the existing guide contract test to protect the completion requirement in both entry points.

Fixes #467

### DIFF
--- a/packages/stim-cli/src/__tests__/guide.test.ts
+++ b/packages/stim-cli/src/__tests__/guide.test.ts
@@ -219,11 +219,14 @@ test('the errors topic documents every code the build commands and the iOS signi
   }
 });
 
-test('current guides expose warm and remove while leaving worktree creation to Git', () => {
+test('current guides require completed warming and leave worktree creation to Git', () => {
   const body = allBodies().join('\n');
   expect(body).toContain('git worktree add');
   expect(body).toContain('stim worktree warm');
   expect(body).toContain('stim worktree remove');
+  for (const guide of [renderTopic('agent'), renderSection('lifecycle', 'options')]) {
+    expect(guide).toContain('Wait for warm to exit successfully (exit code 0)');
+  }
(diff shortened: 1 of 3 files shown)


janicduplessis:
Independent review is clear at c9630bcd0c. Reviewer independently verified all 30 guide tests and both built guide entry points. All local repository checks passed: 3,617 tests and 66 e2e tests, plus formatting, lint, build, typecheck, runtime-floor and knip.

PR #478, opened 2026-09-07
fix: inspect macOS processes when ps is blocked

### LINKED ISSUE #477: Inspect owned macOS processes when sandbox blocks ps
## Problem

On macOS, a sandbox that refuses execution of setuid tools can prevent Stim from inspecting its own collector processes. Cleanup safely refuses to signal an unverified collector, but cannot finish even when the collector belongs to the workspace.

## Evidence

`src/process-args.ts` reads commands and start times through `ps`. A synthetic child-process probe under a macOS sandbox returns no command/start identity when `/bin/ps` cannot execute. A read-only native probe using `proc_pidinfo(PROC_PIDTBSDINFO)` and `sysctl(KERN_PROCARGS2)` can inspect the same child under that policy.

## Cause

Process ownership currently depends on executing `ps`, rather than using the non-privileged native process APIs available to a same-user process. `collector/ownership.ts` correctly treats failed inspection as unverified.

## Fix idea

Add a bounded macOS native inspection path with exact argument boundaries and stable start-time validation. Preserve fail-closed ownership, PID-reuse checks, and existing behavior on other platforms. Verify live, exited, zombie, denied, malformed, and changed-identity cases, plus real sandboxed collector cleanup. Package the helper reproducibly without requiring a global installation or weakening the sandbox.

janicduplessis:
## Description

`stim stop` cannot verify its own log collector when a macOS sandbox refuses to execute `ps`. It correctly leaves the process alone, but cleanup remains incomplete.

Fixes #477.

## Solution

Fall back to read-only macOS process APIs when `ps` execution fails. The helper checks same-user identity before and after reading bounded arguments, includes zombie state, and preserves argument boundaries. Other platforms and successful `ps` calls are unchanged.

The helper is compiled on first use with Apple command-line tools and cached under `STIM_HOME` by source hash and architecture. Compilation or inspection failure remains unverified, never permission to signal a process. This adds a bounded cold-start compilation cost only on the fallback path.

## Test plan

Run `node --test test/e2e/macos-process.e2e.js` on macOS. The real sandbox test covers collector ownership, workspace paths with spaces, exited/zombie processes, protected-file denial, and denied native inspection. The native harness also checks identity changes and argument/environment boundaries.

### DIFF
--- a/packages/stim-cli/shim/macos-process.c
+++ b/packages/stim-cli/shim/macos-process.c
@@ -27,15 +27,20 @@ static bool same_identity(const struct proc_bsdinfo *a, const struct proc_bsdinf
     && (a->pbi_status == SZOMB) == (b->pbi_status == SZOMB);
 }
 
-static bool args_bounds(char *buffer, size_t length, char **start, char **finish, int *argc) {
+static bool args_bounds(char *buffer, size_t length, size_t pointer_size, char **start, char **finish, int *argc) {
   if (length <= sizeof(int)) return false;
   memcpy(argc, buffer, sizeof(int));
   if (*argc <= 0 || *argc > MAX_ARGS_BYTES) return false;
   char *cursor = buffer + sizeof(int), *end = buffer + length;
   size_t size = strnlen(cursor, (size_t)(end - cursor));
   if (size == 0 || size == (size_t)(end - cursor)) return false;
-  cursor += size;
-  while (cursor < end && *cursor == 0) cursor++;
+  /* XNU exec_extract_strings aligns executable_path= plus the path to the target pointer size. */
+  const size_t prefix = sizeof("executable_path=") - 1;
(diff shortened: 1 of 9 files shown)


janicduplessis:
Superseded approach: the user requested persisted unique-pid identities and leaving externally started Metro processes alone. unique-pid PR appandflow/unique-pid#1 is merged with all platform and packaging CI passing; publication is pending. Replacement Stim implementation is in the isolated fix/persisted-process-477 worktree, not yet committed: host ps execution is removed and the real-child collector ownership regression passes, but broader test updates and validation remain. Do not merge this compile-on-first-use helper implementation.

janicduplessis:
Superseded by #487, which uses the published unique-pid library and persisted Stim ownership records. The replacement intentionally leaves externally started processes alone and removes the compile-at-first-use inspection approach from this proposal.

Issue #469, opened 2026-09-07
Benchmark iOS isolation needs native toolchain and recording compatibility checks


janicduplessis:
## Problem

The benchmark's macOS sandbox-exec isolation passes its filesystem-access probes but is incompatible with iOS build steps that apply a second sandbox. A clean SDK 58 ExpoModulesJSI XCFramework build invokes SwiftPM through nested xcodebuild; package resolution fails with `sandbox-exec: sandbox_apply: Operation not permitted`.

The iOS runner preflight should establish that ordinary native build and recording operations work under the actual isolation profile before accepting timed runs. Failed attempts and private benchmark transcripts are intentionally not included here.

## Evidence

- Current origin/main includes the runner profile in `scripts/agent-benchmark/runner-isolation.mjs`.
- A harmless reproduction under a generated benchmark profile is `/usr/bin/sandbox-exec -f <profile> /usr/bin/sandbox-exec -p '(version 1)(allow default)' /usr/bin/true`: exit 71 with the refusal above.
- The SDK 58 installed ExpoModulesJSI `apple/scripts/build-xcframework.sh` calls xcodebuild against a Swift package after clearing stale per-checkout build state. That nested invocation reports the same refusal.
- Native launch artifact preparation alone does not exercise this path if an existing artifact is reused.
- Recording smoke coverage is also needed: agent-device 0.20.10 can report `simctl recordVideo did not expose a complete process identity` in this environment. Its cause needs a separate focused reproduction; it is not yet proven to be the same sandbox limitation.

## Cause

The outer macOS sandbox prevents a child from applying another sandbox. The current probes cover direct/child file access, symlinks, and writes, not native toolchain compatibility. The recording failure's cause is not established.

## Fix idea

Add real untimed iOS toolchain and recording smoke checks under the exact runner policy, and choose a compatible isolation/provisioning design. Preserve forbidden-path denial, per-run device/session ownership, version pins, and identical arm setup. Do not solve this by removing isolation, silently trusting the agent, or relabeling failed timing results as valid. Do not change the frozen campaign until the replacement is reviewed and validated.

janicduplessis:
Claimed: fix/benchmark-ios-compatibility-469

janicduplessis:
Claimed: @janic/issue-469-ios-isolation-smoke

janicduplessis:
[...]

PR #483, opened 2026-09-07
fix: pin native benchmark compatibility adapters

janicduplessis:
## Description

The isolated iOS recovery benchmark cannot finish recording when agent-device's process inspection is denied, and ExpoModulesJSI's nested SwiftPM sandbox blocks its build. Refs #469; the real campaign build/recording/cleanup preflight remains outstanding before timed runs.

## Solution

Prepare a hash-pinned local copy of agent-device 0.20.10 with the native process helper from [the upstream fix](https://github.com/callstack/agent-device/pull/2377), plus an explicit Xcode/JSI sandbox adapter. Both arms use the same copies; the outer benchmark isolation stays enforced. Unknown package bytes or later mutations fail validation.

This is benchmark-only compatibility, not a global install change. It adds a Node process per process query; new campaigns require fresh golden state and exact-toolchain preflight.

## Test plan

- The optional real macOS integration test prepares the published package, proves `ps` is denied while identity lookup works, and verifies Xcode resolves through the pinned wrapper.
- Tampering with package contents, permissions, symlinks, fixture patches, or the executable target fails validation.
- Untimed iOS validation under the existing isolation policy covered build, launch, injected-error capture, Settings recovery, video recording with patched 0.20.10, and cleanup.

### DIFF
--- a/knip.json
+++ b/knip.json
@@ -3,6 +3,7 @@
   "ignore": [
     "website/**",
     "packages/stim-cli/shim/**",
+    "scripts/agent-benchmark/compat/ps-bridge.mjs",
     "scripts/agent-benchmark/stamp.mjs",
     "scripts/agent-benchmark/watch-app.mjs"
   ]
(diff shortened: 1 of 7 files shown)


janicduplessis:
Fresh code re-review is clear at d5c62f152e; all 9 focused live compatibility/isolation tests passed. Campaign use remains gated on fresh golden state and an untimed build, launch-error recovery, recording and cleanup smoke with the exact RC.21 manifest and isolation profile.

PR #484, opened 2026-09-07
feat: support Android launch-error benchmarks

### LINKED ISSUE #482: Run launch-error recovery benchmarks on Android
## Problem

The benchmark driver refuses Android launch-error cells even though Android app-readiness runs and the shared JavaScript root-render fault are supported. The requested matrix needs Stim and control recovery measurements on both platforms.

## Evidence

`scripts/agent-benchmark/driver.mjs` explicitly throws `the launch-crash benchmark currently supports iOS only` during Android dispatch. Crash prompts hard-code `stim ios`, Apple tooling, and simulator names; control fixture preparation only copies iOS ignored outputs.

## Cause

Android readiness support did not extend the launch-crash dispatch, fixture setup, or prompt branches. The diagnosis/recovery parser already accepts a platform parameter, but it needs Android regression coverage through the runnable driver path.

## Fix idea

Enable platform-specific crash setup and instructions while retaining the same injected JavaScript error, isolated worktree/device ownership, initial-error-before-source-inspection rule, and Settings screenshot/video proof. Verify both arms locally before adding timed Android cells. Preserve existing iOS behavior and do not alter historical results.

janicduplessis:
## Description

Android launch-error recovery cannot dispatch, and its watcher would require an APK-label edit that belongs only to the native-change readiness task. Fixes #482.

## Solution

Use the same injected JavaScript failure, runtime-diagnosis requirement, exact source repair, and Settings/video proof on Android. Control creates a fresh pinned AVD and captures logs with its exact serial; Stim uses its owned emulator. Both keep their existing isolation and worktree setup.

The iOS-only guard was introduced with the initial [Android readiness support](https://github.com/appandflow/stim/commit/ea416bf71ff6902453b7ca5ca2adea05faad10af); platform-specific fixture preparation and diagnosis now cover the crash task too. Crash timings remain separate from readiness comparisons.

## Test plan

- Device-level pilot validation is pending: exercise both Android arms before accepting the full matrix.
- Regression coverage accepts fresh-AVD setup and `adb -s emulator-5554 logcat -d` diagnosis, but rejects source inspection before captured runtime errors.
- Verify crash observation does not wait for an APK-label mutation, while native readiness still does.


janicduplessis:
Fresh code re-review is clear at 4665888a1f; required CI is green. Android launch-error Stim/control pilots must retain valid runtime diagnosis, exact repair and same-device Settings/video proof before accepting the full matrix.

janicduplessis:
Merge resolution cc110c91e4 re-reviewed independently: both platform setup and compatibility verification are preserved. 3610 unit tests passed (1 normal opt-in skip),66 E2E and all required local checks passed. Full native pilot qualification still gates timed matrix acceptance.

PR #487, opened 2026-09-07
fix: persist identities for Stim-owned processes

### LINKED ISSUE #477: Inspect owned macOS processes when sandbox blocks ps

janicduplessis:
## Description

Stim cleanup depends on `ps`, which some agent sandboxes cannot execute. It also infers ownership from command names, directories, and ports. This change limits host-process cleanup to instances Stim launched and recorded. Agent-device is unchanged.

## Solution

Persist `unique-pid@0.1.1` tokens for supervisors, collectors, and managed tunnels, and check the exact OS identity before signalling. Cleanup retains live records it cannot verify and leaves replacement processes and their devices alone. External Metro can still be reused but is left running; `stop --force` is removed. Live legacy supervisor records require manual shutdown before reuse.

Exact identities replace the [wall-clock tolerance that protected collectors from clock skew](https://github.com/appandflow/stim/commit/f373e9588afb72984a2bb71c5b911ef75675f0a8), while retaining [caller-group protection](https://github.com/appandflow/stim/commit/5d1d96ac68e7b56016ab0fc83f757b0be4b4b3ae): only a recorded detached supervisor's group is signalled, with no inferred listener fallback.

Denied process inspection retains records; this does not bypass sandbox permissions. The published library's [macOS sandbox fix](https://github.com/appandflow/unique-pid/pull/3) permits capture without a boot ID while retaining the exact start timestamp. Without boot identity, a repeated PID and exact timestamp after a reboot or clock reset cannot be distinguished. Identity checks and signalling remain separate OS calls, not an atomic termination guarantee.

## Test plan

- Real child-process tests cover exact identity, changed/recycled identity, legacy/malformed records, external Metro, and replacement-record cleanup.
- 40 collector-ownership and reclaim tests pass inside the actual Codex sandbox with the published dependency, including replacements registered while cleanup waits for supervisor exit.
- Full suite: 3,602 unit tests and 66 end-to-end tests pass. Unit tests use `npm_config_offline=true npm_config_yes=false` to prevent an unrelated fingerprint fixture's fallback `npx` lookup from waiting on package downloads.

Fixes #477. Replaces #478.


janicduplessis:
Fresh review is clear on aed7aa89df. The reclaim race is covered by real replacement-supervisor and replacement-collector tests; both their processes and device claims survive cleanup. Published unique-pid 0.1.1 is verified inside the actual Codex sandbox. All 3602 unit tests, 66 E2E tests, build, typecheck, lint, formatting, knip and runtime smoke pass. The full unit run uses maxWorkers=4 after a process-start handshake timed out at default concurrency.

Issue #491, opened 2026-09-07
test: avoid implicit React Native downloads in fingerprint fixture


janicduplessis:
## Problem

The benchmark fingerprint integration test can exceed its 60-second timeout because its minimal fixture has no React Native CLI. This delays or fails otherwise passing local checks.

## Evidence

On Node 26.7.0 and Vitest 5, the test in scripts/agent-benchmark/cache-key.test.mjs took 92–96 seconds and timed out both in the full suite and alone. Process inspection showed npm exec react-native config. The full suite passed with npm_config_yes=false npm_config_offline=true (3611 passed, one skipped), without changing assertions or timeouts.

## Cause

benchmarkFingerprint invokes the pinned @expo/fingerprint package, whose getCoreAutolinkingSourcesFromRncCliAsync executes npx react-native config. The test fixture deliberately has no React Native package, allowing npx package resolution outside the fixture.

## Fix idea

Make the integration fixture self-contained or prevent implicit package installation for this test while retaining real fingerprinting and the conflicting-fingerprint-package regression. Do not raise the timeout or change benchmark runtime configuration.

janicduplessis:
Claimed: @janic/issue-491-fingerprint-fixture

PR #492, opened 2026-09-07
fix: clarify benchmark launch-error capture gate

### LINKED ISSUE #490: Clarify delayed-error capture gate in launch-error benchmark prompts
## Problem

The launch-error benchmark tells the Stim agent to run `stim logs --errors` immediately after launch, then allows source inspection after those diagnostic commands. A delayed JavaScript error can arrive after that first query returns empty. The agent can follow that sequence and diagnose the visible redbox, yet fail the audit because the error token was not captured in logs before source inspection.

## Evidence

A local validation sequence reproduced the expected audit rejection: empty log query, device snapshot showing the error, source inspection, then a token-bearing log query. `launchCrashDiagnosis` returns `launch-crash-pre-capture-command-not-allowed`; successful repair and Settings proof do not satisfy the earlier log-first requirement. Campaign transcripts and rejected benchmark results remain machine-local.

## Cause

`scripts/agent-benchmark/launch-crash-setup.mjs` describes completing diagnostic commands, while `launchCrashDiagnosis` in `scripts/launch-crash-benchmark.mjs` requires a completed token-bearing log capture before inspection. The prompt does not explicitly distinguish an empty successful query from captured failure evidence.

## Fix idea

Make the prompt state the existing evidence gate explicitly for both arms: wait/re-query until a completed log command prints the launch error and source location before reading/editing source. An empty query or visible UI error alone does not satisfy the log-first measurement. Preserve validators, targets and prior verdicts. Validate the delayed-error ordering in a regression case, then run a fresh pilot.

janicduplessis:
## Description

A JavaScript launch error can reach the logs after the benchmark's initial query returns empty. The prompt allows agents to mistake that completed query for the evidence required before source inspection. Fixes #490.

## Solution

Tell both arms to wait for a completed log query containing the error and source location before reading or editing source. Keep the existing audit criteria and prior verdicts unchanged.

## Test plan

The delayed-error regression verifies that an empty successful query and a redbox snapshot do not advance diagnosis time; the subsequent token-bearing log capture does. Run a fresh launch-error pilot with the clarified prompt before publishing results.


janicduplessis:
Independent review found that a separate sleep command would fail the existing Stim audit. The prompt now explicitly requests repeated standalone log queries without separate sleep/wait commands. Fresh re-review is clear, including synthetic iOS/Android × Stim/control repeated-query checks. Full local checks pass; unit tests use npm_config_yes=false npm_config_offline=true to avoid the separately tracked fixture issue #491. Existing audit criteria are unchanged.

PR #495, opened 2026-09-07
fix: require benchmark warm completion before use

### LINKED ISSUE #493: fix: audit benchmark warm completion before dependent commands
## Problem

The benchmark setup audit accepts a successful worktree warm even when start or native launch ran before warm completed. A shell-tool wrapper can finish after returning a running session handle; printing only its output hides that handle and can make an agent mistake empty output for completion.

## Evidence

The current benchmarkSetupInvalidReasons only checks whether any warm command eventually exited zero. A command sequence with warm running from t=0 to t=90 and stim start at t=60 therefore passes that check. Published guide agent already requires warm to finish first. Private troubleshooting confirmed the same ordering can temporarily leave Metro dependencies unavailable; no rejected benchmark results are included here.

## Cause

successfulCommand in scripts/agent-benchmark/run-guards.mjs tests exit status but not the warm end event relative to dependent command starts. The common task prompt does not explain shell-session completion versus code-wrapper completion.

## Fix idea

Require warm completion before dependent start/native/install commands using recorded command timestamps/events, with regression cases for overlap and correctly polled completion. Clarify for both arms that code wrappers must preserve complete shell results and poll returned session handles to an exit status; detached-process monitoring remains unchanged. Keep this separate from CLI runtime changes.

janicduplessis:
## Description

The benchmark accepts a successful warm even when Metro started before copying finished. Agents can mistake a code wrapper returning for shell completion when the wrapper hides its running-session handle. Fixes #493.

## Solution

Require recorded warm completion before dependent Stim commands, rejecting overlapping or unproven ordering. Retain unfinished commands in the audit; Claude background submission is not completion, and a matching terminal TaskOutput result supplies the completion evidence. Tell both arms to preserve shell result metadata and poll foreground sessions to an exit status; detached-process monitoring stays unchanged. This changes the harness, not Stim runtime or native cache settings.

## Test plan

Verify that warm running from 0–90s rejects a start at 60s but accepts one at 91s. Regressions also cover equal timestamps with event ordering, unfinished overlapping commands, and Claude task results with missing, mismatched, failed, or successful completion evidence. Run a fresh launch-error pilot with the clarified shell-completion instructions.


janicduplessis:
Independent review is clear after addressing both command-evidence findings; reviewer independently verified 63 parser, guard, and launch-crash tests. Full local format, lint, build, typecheck, knip, unit, E2E, and runtime checks passed. Local unit checks used npm_config_yes=false and npm_config_offline=true to avoid the unrelated implicit npx download tracked in #491.

PR #497, opened 2026-09-08
fix: support managed sessions in launch-error benchmarks

### LINKED ISSUE #496: fix: make launch-error control bootstrap compatible with managed shell sessions
## Problem

The launch-error control prompt requires detached shell processes and forbids long foreground calls. In the current Codex runner, nohup jobs disappear when their launching command returns, sending the agent through process-management workarounds instead of the app task. The prompt also leaves Android AVD storage implicit, although the observer needs the inherited directory to verify the system image.

## Evidence

Observed on current main 39617444d18b05503737065ec92c21d98fd72722 with Codex 0.145.0-alpha.11 and Stim RC.21. An Android control attempt had two vanished copy jobs and vanished emulator jobs; foreground rsync completed, and Ruby Process.daemon eventually kept emulator and Metro alive. The agent moved ANDROID_AVD_HOME to a private tmp directory; the observer found the expected emulator name/model/runtime but could not resolve systemImage, so correctly refused the run. Private failed-run evidence is retained, not published.

A harmless host shell probe reproduces the lifetime difference: launch `nohup /bin/sleep 30` from bash, verify kill -0 inside that command, then check from the next exec_command; the process is already gone. A foreground `/bin/sleep 8` yielded a session ID and completed with exit 0 when polled. This probe uses the desktop shell tool; a minimal same-CLI/sandbox qualification should precede another timed control run.

## Cause

scripts/agent-benchmark/launch-crash-setup.mjs mandates one process-launch mechanism even though the required evidence is a completed standalone error-log capture before source inspection. The exact runner cleanup mechanism is not yet established. Android observation resolves AVD metadata from inherited ANDROID_AVD_HOME, which the prompt does not explicitly preserve.

## Fix idea

Allow managed shell sessions for control with complete result metadata and polling, just as for Stim, while retaining explicit completed error capture before source reads/edits. Keep server sessions alive through proof instead of treating a yielded session as completion. Explicitly preserve inherited Android toolchain/storage variables. Keep device, source-repair, timing and cache validation unchanged; do not make historical invalid attempts valid.

janicduplessis:
## Description

Launch-error control was forced to use detached shell jobs. In a probe using the campaign's Codex CLI and sandbox, a `nohup sleep` child disappeared between shell calls; a managed session completed normally. This changes benchmark guidance and its command audit, not Stim runtime behavior. The runner's exact cleanup mechanism is still unconfirmed.

## Solution

Allow managed control sessions, keeping servers alive through readiness and proof while polling finite jobs to completion. Accept direct local Expo Metro startup in the audit without accepting compound source reads or unfinished native-launch evidence. Explicitly preserve inherited Android storage paths so the observer can verify the emulator's system image.

The [original detached-only rule](https://github.com/appandflow/stim/blob/c4fdbca0b60705e94baf3dd32d83c7a6944bf937/scripts/agent-benchmark/launch-crash-setup.mjs#L29) accompanied separate error capture; that requirement remains but does not need detached jobs. Historical invalid attempts remain invalid; the next control pilot has not run yet.

## Test plan

- Same-CLI sandbox probe: detached child vanished between calls; managed session completed; inherited Android paths were intact.
- Regression cases on both platforms accept live Metro plus completed launch/error capture, and reject unfinished launch, source reads and command substitutions.
- After merge, retry the failed Sol Android control cell with existing version/cache/device/proof gates unchanged.

Fixes #496


janicduplessis:
Independent fresh-agent review is clear at f5dff02fedaeb87d63713f6d836e2591a24d97db. Reviewer inspected issue, diff, surrounding audit and guidance, and independently passed 33 affected tests. Finite launch completion and standalone error capture remain enforced; live managed Metro is accepted. All required local checks also passed (3630 unit, 66 E2E and runtime). Timed control pilot remains pending until merge; historical invalid evidence is retained.

PR #499, opened 2026-09-08
fix: recognize scoped setup in launch-error audit

### LINKED ISSUE #498: fix: recognize scoped launch-error setup commands in benchmark audit
## Problem

The launch-error audit rejects ordinary control setup commands before error capture. A managed-session Android control run built, diagnosed the injected JS error before reading source, made the exact repair, and saved Settings/video proof, but was rejected for its finite dependency-copy script starting with `set -e`.

## Evidence

Current main e1f9a15b49b1324f467f77db3e8f15e1cceadd13, campaign RC.21, Sol Android control attempt1788828516048. Native build2m24s (304 executed/303 cache), Settings538.538s, matching emulator metadata, isolated proof and recording passed. The normal verdict is invalid with `launch-crash-pre-capture-command-not-allowed` at item_4; diagnosis/usage failures cascade from it. Private evidence is retained; verdict unchanged.

A read-only diagnostic sweep identified additional legitimate setup forms the allowlist rejects: quoted `$ANDROID_HOME` executable paths, `printenv ANDROID_AVD_HOME`, reading the owned AVD config and emulator discovery INI, `tee` log pipelines, architecture assignment before Expo, and the required AVD config edit made with apply_patch. The completed token-bearing log query precedes app source inspection. Existing validation stops at only the first rejection.

## Cause

`scripts/launch-crash-benchmark.mjs` matches command spelling rather than consistently recognizing permitted setup operations. `topLevelShellCommand` in run-guards.mjs strips only one outer quote pair and does not decode concatenated shell-quoted fragments in Codex's rendered commands. The current allowlist also accepts some broad prefixes (for loops/nohup) while rejecting equivalent transparent forms.

## Fix idea

Normalize shell wrappers without execution and classify scoped setup operations, transparent log pipelines, SDK tool invocations and exact owned AVD metadata edits. Keep unknown commands, app/source inspection, hidden command substitution and unproven launch/error-capture completion rejected. Report all setup violations for diagnosis rather than requiring another timed attempt per spelling. Exercise a full synthetic control setup through the audit, including negative source-access cases, before another timed retry. Do not change runtime/cache/device pins or publish historical invalid attempts.

janicduplessis:
## Description

Launch-error control can recover the app correctly but fail its audit on ordinary setup syntax: `set -e` copy loops, quoted SDK paths, and `tee` logging. Codex's concatenated shell quotes also expose a decoding bug in normalization shared by other benchmark audits.

The [original pre-capture allowlist](https://github.com/appandflow/stim/commit/a19b03a4476c9490c162275541e2df4f0980c039) deliberately prevents source inspection before diagnosis. This keeps that rule while recognizing more of the setup the benchmark itself requests.

## Solution

Decode one shell-wrapper layer without execution, recognize scoped control setup, and report every rejected operation instead of only the first. AVD edits must target the configuration path recorded at dispatch; the copy-loop allowance is limited to recognized native/dependency inputs and the run worktree.

A launch piped through `tee` counts only when that shell enables `pipefail`; otherwise zero exit status proves only the log writer succeeded. The control prompt requires this explicitly.

This is benchmark-only. Unknown operations remain rejected, and filesystem isolation remains the security boundary. Historical invalid results stay unchanged; the timed retry has not run yet.

## Test plan

- Replay retained control evidence read-only: setup forms are recognized, but its `tee` launches lack upstream exit evidence and remain unqualified.
- Exercise a complete synthetic managed Android setup plus rejection cases for hidden source reads, substitutions, another AVD, and multiple violations. Existing shell quoting and proof-boundary cases still pass.
- After merge, retry Sol Android control with version, cache, device, timing and proof gates unchanged.

Fixes #498

### DIFF
--- a/docs/agent-benchmark.md
+++ b/docs/agent-benchmark.md
@@ -144,6 +144,9 @@ partition, system image, and default Quick Boot policy as Stim.
 Launch-error control may use runner-managed sessions instead of detached shell
 jobs and keeps the inherited Android SDK, AVD, Gradle, and emulator-report
 locations unchanged so device verification and cleanup see the same metadata.
+The launch-error audit recognizes scoped setup and log capture across shell
+quoting forms. It reports all rejected operations for investigation; source
+inspection before completed error capture still invalidates the attempt.
 Installing dependencies inside the timed interval, a failed `guide agent` or
 `worktree warm`, and an Android native build without Stim's `--build-cache`
 evidence invalidate the attempt.
--- a/scripts/agent-benchmark/README.md
+++ b/scripts/agent-benchmark/README.md
@@ -184,6 +184,13 @@ for the separate completed error query and preserves the inherited Android SDK,
 AVD, Gradle, and emulator-report locations so observation and cleanup use the
 same metadata as the agent.
 
+The pre-capture audit decodes shell quoting without executing it and recognizes
+scoped setup operations, including local SDK tools, managed log pipelines, and
+the assigned AVD's configuration edit. Collection reports every rejected command
+in `diagnosis.violations`, retaining the first rejection in `commandId`. Unknown
+commands and source inspection still invalidate the attempt; this is not a
+replacement for the runner's filesystem isolation.
+
 A launch can finish before the JavaScript error reaches its log. Both arms must
 repeat standalone foreground log queries, without separate sleep or wait commands,
 until a completed query prints the error and source location before inspecting or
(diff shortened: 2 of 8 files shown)


janicduplessis:
Fresh independent review is clear after resolving both findings: metadata reads now accept only a single scoped INI target, and tee-wrapped launches require same-shell pipefail before they count as successful. Reviewer independently passed 62 tests and six real zsh pipeline cases. Full local format/lint/build/typecheck/knip/unit/E2E/runtime checks passed again. Initial Node22 CI timeout passed at identical commit on retry and is tracked separately in #500. Waiting for CI on final head before merge.

PR #513, opened 2026-09-08
fix: review harmless launch-error benchmark deviations

### LINKED ISSUE #512: fix: review non-behavioral launch-error benchmark deviations
## Problem
A completed launch-error control attempt with valid diagnosis, exact source repair, Settings screenshot and recording was rejected for a CocoaPods checksum-only lockfile update and a separate run-scoped diagnostic session. Unscoped help queries also trigger the session guard despite not opening a session.

## Evidence
The only additional tracked change is one ExpoModulesCore SPEC CHECKSUMS value in ios/Podfile.lock; dependency versions and the rest of the lockfile are unchanged. The diagnostic session uses the exact unique run ID plus a suffix, targets the same assigned simulator, and closes before the required proof session. Raw evidence and original verdict are retained privately.

## Cause
The source audit treats every extra tracked path as a behavioral change. Session validation compares a literal command prefix and does not distinguish help from device mutation or a safely scoped auxiliary session.

## Fix idea
Retain integrity-bound review evidence for non-behavioral deviations rather than rerunning a successful attempt. Narrowly validate checksum-only lockfile changes and closed auxiliary sessions bound to the same run, state directory and device. Keep rejecting changed dependencies/source, default or foreign sessions, wrong devices, missing repair/proof and daemon interference. Preserve original run records and require reviewed evidence before publication.

janicduplessis:
## Description

A Sonnet iOS control run repaired the launch crash and reached Settings in 13m 22s, but validation rejected a checksum-only `Podfile.lock` update and a closed diagnostic session named `<runId>-diag`. Fixes #512.

The [source-only rule](https://github.com/appandflow/stim/commit/a19b03a4476c9490c162275541e2df4f0980c039) and [literal session guard](https://github.com/appandflow/stim/commit/782b8c5046cf300fb54dddf509ca9240cdb61cf5) intentionally enforce repair scope and isolation; these cases preserve that intent without matching the syntax.

## Solution

Allow checksum-only CocoaPods updates and same-run, same-device diagnostic sessions that close before the proof session, with explicit review required for either arm. Help queries need no session. Retained evidence is hash-checked and revalidated before publication; original verdicts remain untouched.

This changes benchmark acceptance, not Stim runtime behavior. A mistaken exemption could admit an incomparable run; wrong/default sessions, changed dependencies or source, missing proof, and daemon interference still fail.

## Test plan

Re-exported the retained Sonnet run locally: its existing diagnosis and Settings proof are accepted without another timed attempt. Regression cases reject changed dependency versions, wrong devices, missing session closure, and altered or incomplete review evidence.

### DIFF
--- a/scripts/agent-benchmark/README.md
+++ b/scripts/agent-benchmark/README.md
@@ -256,7 +256,20 @@ unrecognized setup command. It contains `schemaVersion: 1`, `runId`,
 `shellParserSha256` (run-guards), and ordered `commands` entries with `commandId`,
 the exact `command`, and a nonempty `assessment`. Export re-derives diagnosis,
 recovery and diagnosis-time usage from hash-verified retained evidence. A review
-can clear only setup-syntax rejection and its missing-diagnosis/usage consequences;
-it cannot clear other failure reasons or source-before-capture violations. Valid
-runs with setup warnings also require review before publication. The original
+can clear setup-syntax rejection and its missing-diagnosis/usage consequences.
+Read-only agent-device help does not need a session. An auxiliary diagnostic
+session must use the exact run namespace and state directory, open the assigned
+device, and close successfully before the pinned proof session opens. Its
+`auxiliarySessions` review entries name each `session` and a nonempty `assessment`.
+Default or foreign sessions, device overrides, missing closure and daemon
+interference remain failures.
+
+An iOS `Podfile.lock` update is reviewable only when every difference is a valid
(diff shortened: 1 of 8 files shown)


janicduplessis:
Fresh independent review is clear after addressing command-overlap and quoted-argument cases. The 101 focused regressions and all repository checks pass, including native integration and 66 end-to-end tests. Re-derived the retained event graphs for all twelve completed workflows with unchanged timings and reviewed command sets; originals remain untouched. Waiting for final exact-head CI before merge.

PR #515, opened 2026-09-08
fix: recognize completed benchmark launch pipelines

### LINKED ISSUE #514: fix: recognize completed benchmark launch pipelines
## Problem

A successful Opus Android launch-error control run was rejected as missing initial launch evidence because it enabled pipefail after changing directory. The only hard reasons are initial-launch-evidence-missing and its diagnosis consequence; the exact source repair, Settings screenshot and recording passed.

## Evidence

The agent ran `cd <worktree> && set -o pipefail && npx expo run:android ... | tee <log> | tail -40; echo "PIPELINE_EXIT=$?"`. Captured output proves `BUILD SUCCESSFUL in 2m 38s`, installation/opening on its exact emulator, and `PIPELINE_EXIT=0`. A separate later logcat query captured the injected error before any source inspection. Settings proof completed in 462.013s. Original records and media remain private and unchanged.

## Cause

`successfulLaunch` requires the shell command to begin with `set ... pipefail` followed by semicolon/newline. It does not recognize a preceding `cd` or an AND-list separator, and must also account for the trailing exit-status echo instead of trusting the shell's final echo exit code.

## Fix idea

Recognize scoped directory setup and pipefail-enabled launch pipelines, validate an explicit zero pipeline status when a trailing echo masks the shell status, and retain rejection of disabled/missing pipefail and nonzero pipeline status. Let the existing evidence-bound review recover this completed run without changing original records or rerunning it.

janicduplessis:
## Description

Opus repaired the Android launch crash and reached Settings in 7m 42s, but the validator rejected `cd <worktree> && set -o pipefail && ... | tee <log>; echo "PIPELINE_EXIT=$?"`. It only recognized pipefail at the start of the command with a semicolon or newline. Fixes #514.

The [original pipeline check](https://github.com/appandflow/stim/commit/44fed2eb5f42e2d589e9cfc6df5df954cb6ae4d7) prevents `tee` from masking launch failure; this retains that intent while accepting directory setup and explicit pipeline-status reports.

## Solution

Recognize those completed pipelines and require exactly one zero status report when a trailing echo masks the shell exit. Missing/disabled pipefail and failed or ambiguous status reports still fail.

Previously rejected runs can pass review when their recorded evidence revalidates; original records remain unchanged. No Stim runtime change.

## Test plan

Re-derived diagnosis for the retained Opus run (6m 23s) and verified all 12 earlier runs retain their event graphs and timings. Real Bash regression cases both exit zero at the shell level but accept only the successful upstream pipeline. Export regression accepts reviewed retained evidence and still rejects missing launch evidence.

### DIFF
--- a/scripts/agent-benchmark/README.md
+++ b/scripts/agent-benchmark/README.md
@@ -207,6 +207,11 @@ shell programs are safe or replace the runner's filesystem isolation.
 Build/launch pipelines ending in `tee` must enable `set -o pipefail` in the same
 shell command. Without it, zero exit status proves only the log writer finished,
 so the command cannot establish initial launch success.
+The pipeline may follow `cd <worktree> && set -o pipefail && ...`. If it ends
+with `; echo "PIPELINE_EXIT=$?"`, the captured output must contain exactly one
+`PIPELINE_EXIT=0` line: the echo's own zero exit does not prove build success.
+Retained launch-evidence rejections can be reviewed only when re-derived commands
+prove successful launch, separate error capture, repair and Settings proof.
 
 A launch can finish before the JavaScript error reaches its log. Both arms must
 repeat standalone foreground log queries, without separate sleep or wait commands,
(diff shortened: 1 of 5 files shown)


janicduplessis:
Fresh independent review is clear after covering failed directory setup: a skipped pipefail command must not let tee mask an upstream failure. All repository checks passed, including native integration and 66 end-to-end tests; 72 focused launch/export tests passed with real Bash status-masking cases. Retained evidence for all thirteen completed workflows re-derives with unchanged timings. Waiting for final exact-head CI before merge.

PR #517, opened 2026-09-08
fix: accept benchmark command exit reports

### LINKED ISSUE #516: fix: accept successful benchmark commands with exit status reports
## Problem
Successful launch-error runs are rejected when commands append `; echo "EXIT=$?"`. Opus Android Stim reached Settings in 284.889 seconds but was marked missing warm and error capture.

## Evidence
Retained rc.21 run: opus-android-launch-error-opus-launch-crash-stim-1788848772227. Warm completed with 0 failed entries and EXIT=0 before start; a later stim logs --errors printed the injected token and RootLayout frame before source inspection. Source repair, Settings screenshot, recording, runtime version and artifact hit passed.

## Cause
On current main ca7df6cd2d, successfulCommand only accepts a final Stim segment, while errorCaptureCommand requires whitespace after --errors. The reporting echo becomes the final segment and the semicolon fails the log regex.

## Fix idea
Recognize a trailing exit-status report and validate its captured status. Reconstruct setup ordering and diagnosis from retained events before accepting a reviewed result. Preserve original records and all isolation, repair and proof checks; do not rerun a successful workflow for harmless shell formatting.

janicduplessis:
## Description

Opus repaired the Android launch crash and reached Settings in 4m 45s, but the benchmark rejected `stim worktree warm; echo "EXIT=$?"` and `stim logs --errors; echo "EXIT=$?"` as missing steps. Fixes #516.

## Solution

Accept these status-reporting commands using the captured status, not the echo's exit code. Missing, duplicated or nonzero reports cannot establish success. Retained runs can be reviewed without rerunning them or rewriting their original records; warm ordering and runtime diagnosis must still be proven from the events. This changes benchmark validation only.

## Test plan

- Real Bash commands returning 0 and 7 both end with a successful echo; only the actual success passes. Overlapping warm still fails.
- Reconstructed all 14 retained workflows: existing diagnosis/recovery evidence remains valid, including the previously rejected Opus run (3m 57s diagnosis, 4m 45s Settings).
- Export regression accepts reviewed warm/log false rejections while rejecting failed warm and preserving the original record.

### DIFF
--- a/scripts/agent-benchmark/README.md
+++ b/scripts/agent-benchmark/README.md
@@ -212,6 +212,10 @@ with `; echo "PIPELINE_EXIT=$?"`, the captured output must contain exactly one
 `PIPELINE_EXIT=0` line: the echo's own zero exit does not prove build success.
 Retained launch-evidence rejections can be reviewed only when re-derived commands
 prove successful launch, separate error capture, repair and Settings proof.
+Ordinary commands may also append `; echo "EXIT=$?"`. The audit uses the single
+captured status line, not the echo's exit code; missing or ambiguous reports do
+not prove success. A retained warm-status rejection requires re-derived setup
+checks, including warm completion before start or build, before publication.
 
 A launch can finish before the JavaScript error reaches its log. Both arms must
 repeat standalone foreground log queries, without separate sleep or wait commands,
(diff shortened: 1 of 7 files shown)


janicduplessis:
Independent review is clear after addressing exit masking and preserving existing validated tee pipelines. Real Bash regressions cover successful/failed reports, intervening true commands, and both Stim/control pipeline handling. All full local checks passed; the fingerprint fixture uses the documented offline/no-install environment for existing issue491. Original benchmark evidence remains untouched.

PR #529, opened 2026-09-08
fix: clean benchmark exit-wrapper commands

### LINKED ISSUE #528: Clean benchmark directory prefixes with explicit exit markers
## Problem

The published benchmark still shows repeated literal `cd <worktree> &&` prefixes for commands wrapped with `; echo "EXIT=$?"`, despite the command-presentation cleanup in #521.

## Evidence

Current origin/main 0912522d7d contains 27 commands whose display still starts with `cd`. Opus Android launch-error includes `cd ./worktrees/bench-launch-crash-stim && stim worktree warm; echo "EXIT=$?"`, with a captured successful EXIT marker.

https://appandflow.github.io/stim/benchmarks/details?benchmark=opus-android-launch-error&run=launch-crash-stim

## Cause

The presentation sanitizer correctly refuses to infer successful directory setup from a shell exit code masked by a following semicolon command. It does not yet recognize the harness/agent's explicit exit-status echo wrapper.

## Fix idea

Recognize a narrow trailing exit-status echo and use its captured result before presenting a literal directory prefix as context. Preserve raw sanitized commands and evidence, do not hide failed directory setup, and update current public datasets plus future export processing. Test missing/failed/ambiguous markers as well as successful wrappers.

janicduplessis:
## Description

Some benchmark commands still show their directory prefix because `; echo "EXIT=$?"` masks the command's shell exit status. Recognize that narrow wrapper only when its captured marker reports success, keeping failed or ambiguous setup visible.

Fixes #528.

## Solution

Use the captured exit marker when deriving display context, and refresh the eight affected Opus Android launch-error commands. Original commands, output, timing, and validity remain unchanged; future exports use the same logic.

## Test plan

Open Opus / Android / Launch crash / Stim and select `stim worktree warm`, `stim start`, or `stim android`. The terminal and timeline show the command without `cd` or the exit-status echo; expanding command context still shows the complete original. Failure and ambiguous-marker regression cases retain their original directory setup.

The existing fingerprint fixture still exceeds its local 60-second timeout; the other 3,653 unit tests and all 66 end-to-end tests passed.

Before:

Before

After:

After

### DIFF
--- a/scripts/benchmark-command-presentation.mjs
+++ b/scripts/benchmark-command-presentation.mjs
@@ -38,9 +38,16 @@ function shellTokens(command) {
   return tokens;
 }
 
-export function benchmarkCommandPresentation(command, exitCode) {
+export function benchmarkCommandPresentation(command, exitCode, output = '') {
   if (command.includes('<<')) return undefined;
   let display = command;
+  const statusEcho = display.match(/;\s*echo\s+"(EXIT|PIPELINE_EXIT)=\$\?"\s*$/);
+  if (exitCode === 0 && statusEcho) {
+    const markers = output.split(/\r?\n/).filter((line) => line.startsWith(`${statusEcho[1]}=`));
+    if (markers.length === 1 && markers[0] === `${statusEcho[1]}=0`) {
+      display = display.slice(0, statusEcho.index);
+    }
+  }
   let cwd;
   let tokens = shellTokens(display);
   if (!tokens) return undefined;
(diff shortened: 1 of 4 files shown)


janicduplessis:
Independent review: CLEAR at 8fff6b8f9d. Verified failed/missing/duplicate/ambiguous exit markers retain directory setup, all 1,905 published command records match the presentation function, and only eight records gain display metadata. Independent parser/exporter/timeline checks: 51 tests passed. Exact-head CI passed.

PR #540, opened 2026-09-08
feat: publish activity-timed readiness benchmarks

### LINKED ISSUE #539: Publish Android readiness-integration benchmark checks
## Problem
The public benchmarks do not include the completed Android startup-error reruns with optional readiness logs and the merged launch-timing fixes.

## Evidence
The September 8 campaign has validated Sol, Luna and Opus Stim runs with initial-command error capture, artifact hits and Settings proof. Sonnet recovered but skipped the required guide; retain its raw invalid audit and exclude it from published results. Claude runs also include approximately 95 seconds before runner initialization.

## Cause
The website still contains the earlier matched fixture comparisons. The new Stim-only readiness fixture has no freshly matched control runs.

## Fix idea
Publish the three valid runs as a separate readiness-integration check with command audits, proof, full usage/cost and startup-timing caveats. Keep existing matched comparisons intact; do not infer new speedups or change audit validity.

janicduplessis:
## Description

Benchmark times include runner startup: Opus's latest Android run took 4m49s from dispatch, but 3m13s from its first recorded activity. The website now starts every Stim and control clock at the first recorded agent message or shell command, retaining the excluded offset and dispatch totals. This excludes any unobserved initial reasoning too; it is not prompt-to-result latency.

## Solution

Most of the diff is mechanical timestamp conversion in published JSON. Preserve raw evidence, command durations, full-turn cost and recordings; use the same clock policy for future exports.

Publish the latest validated Android readiness-log checks separately, with their own audit links. Their fixture changed and controls were not rerun, so they must not replace matched comparisons.

## Test plan

Verify that old comparison links still open the same runs, timelines start at zero, command durations stay unchanged, and each audit shows its excluded lead-in. Open the latest readiness checks to inspect cache-hit/error output, Settings proof, video and full-turn cost. The existing 49 runs were compared against a deterministic timing-only conversion of the original data.

Fixes #539

readiness-site-539-sonnet


janicduplessis:
Verification: all 49 existing runs across 17 datasets exactly match a deterministic timing-only conversion of origin/main. Browser-checked overview, Opus and Sonnet deep links, corrected clocks and full-turn cost; UI screenshot attached above. Focused benchmark tests: 78 passing. E2E: 66 passing. Website build/typecheck, format, lint, package build/typecheck and knip pass. Full local suite: 3,703 pass, one skipped, with only the existing fingerprint-fixture timeout tracked in #491. Exact-commit CI passes on Node 22 and 24 plus published-runtime Node 20.19.4. Mobile viewport override was blocked by tool policy, so no new mobile-specific visual check is claimed.

janicduplessis:
Fresh independent review is clear at b0141384434e3e330a1f744c4473375c0ec1ddd0: all 49 conversions independently verified, all 94 proof/video assets present, four new runs cross-checked against raw timestamps and byte-identical evidence. Initial launches show artifact cache hits and injected errors; the original invalid Sonnet attempt remains excluded and the separate retry is valid. No actionable findings.

PR #548, opened 2026-09-09
fix: replace superseded readiness benchmarks

### LINKED ISSUE #546: Publish one current readiness benchmark per model and platform
## Problem

Benchmark links still show an older Android launch that missed the startup error, even though newer readiness-log runs catch it. Separate integration-check entries make it unclear which result is current.

## Evidence

The Luna `luna-android-launch-error` audit returns before the injected error; `luna-android-readiness-entry-error` captures it in the initial platform command. PR #540 intentionally kept these separate. The user now requests replacement in place, retaining existing controls without optional readiness logs, and readiness coverage for every model on both platforms.

## Cause

The catalog keeps the latest four runs outside the primary comparison list because they use an updated readiness-log fixture.

## Fix idea

Publish one current run per model/platform/scenario/arm. Replace superseded Stim launch-crash results, route previous links to the current canonical benchmark, remove duplicate sections, and disclose the optional-integration difference. Preserve raw evidence privately. Verify existing iOS support and run the four iOS models with readiness logs before publishing replacements.

janicduplessis:
## Description

Existing benchmark links still showed launches that missed the startup error, while newer readiness-aware runs lived under separate entries. Reverse the separation introduced in [#540](https://github.com/appandflow/stim/commit/1f9fc1e87e5f15e3b215ea8fcd9236db8a60980b) at the user's request: old links show the latest validated result rather than a second benchmark.

## Solution

Replace all four models' iOS and Android Stim launch-error arms with their validated readiness-log reruns and keep existing controls. The comparison explicitly identifies the optional logs on Stim and their absence on control; this is integrated Stim versus the standard toolchain, not identical instrumentation. Show per-arm dates and always publish the latest validated run. The exporter selects the newest valid dispatch within each stage, keeps stable arm IDs, and rejects mixed-model/platform stages rather than collapsing different benchmark cells.

Old integration-check and earlier Sol links resolve to canonical benchmarks. Raw campaign evidence remains unchanged; remove only superseded public datasets, retaining the recordings used by current results.

## Test plan

Open the existing Luna Android launch-error link: its initial `stim android` output now includes the readiness wait and injected error. Check the equivalent initial `stim ios` output for all four models. Switch to control to verify the retained timeline, then switch model/platform/scenario. Old integration-check links resolve to the same current pair; no duplicate catalog entries remain. Inspect per-arm recording dates and readiness labels.

Fixes #546

Related: #557 makes the existing guide prerequisite explicit for the new Sonnet and Opus iOS runs.

stim546-replacement


janicduplessis:
Final independent review is clear for the exporter, all eight readiness replacements, controls, aliases, proof images/videos, and the data-independent usage display fixtures. Each current Stim launch-error run hits its artifact cache, observes the readiness wait, reports the injected error from the initial platform command, and finishes with Settings proof. All repository checks ran locally: format, lint, build, typecheck, knip, 66 end-to-end tests, runtime checks, and website typecheck/build passed. The full unit run found the old mutable Opus-count assertion (fixed; all 60 affected tests now pass) plus the already-tracked unrelated fingerprint timeout #491. Waiting for exact-head CI before merge.

janicduplessis:
Published: Docs deployment 34304830347 succeeded for merge adba239674367494173a1bd700df502ae4301844. Browser-verified the original Luna Android launch-error URL and switched to the new Luna iOS result. Both initial platform commands show the readiness wait, artifact cache hit, and injected launch error. The existing links now expose canonical replacements, with retained controls and per-arm readiness labels.

PR #550, opened 2026-09-09
test: fix the intermittent supervisor handoff timeout

### LINKED ISSUE #500: test: investigate intermittent supervisor handoff timeout on Node22
## Problem

The Node22 CI unit suite intermittently times out in the supervisor handoff cleanup test, then a later test fails restoring its working directory because the timed-out test's temporary directory has been removed.

## Evidence

[CI run34175922752](https://github.com/appandflow/stim/actions/runs/34175922752), commit277d3cd0c75cf5dfc5c1d7d5ab7ef1bf42d48f5c, attempt1: `start.test.ts:1315` (a supervisor handoff write failure reaps a child that ignores SIGTERM) exceeded5000ms; `start.test.ts:1407` subsequently hit ENOENT in `runConcurrentActions` restoring cwd atline202. Node24 and local macOS passed. Rerunning only the failed job at the same commit passed on attempt2. The PR changes benchmark scripts, not start or this test.

## Cause

The timing failure is not diagnosed. The observed secondary cwd error follows timeout cleanup. The test already injects termination timing and a fake child, so simply increasing its timeout may hide the actual issue.

## Fix idea

Reproduce on Linux Node22 with CI-like load, identify any real wait left in the test, and ensure timeout cleanup cannot race with later tests' process-wide cwd restoration. Preserve the SIGTERM-to-SIGKILL behavior assertion.

janicduplessis:
## Description

`start.test.ts > a supervisor handoff write failure reaps a child that ignores SIGTERM` intermittently times out on the Node 22 CI job, and the next test then fails with `ENOENT ... chdir` while restoring its cwd ([run 34175922752, attempt 1](https://github.com/appandflow/stim/actions/runs/34175922752/attempts/1)).

The timeout is a pid collision, not load. The fake child used `pid: 4242`; when a real process holds that pid on the runner, [`waitForSupervisorHandoff`](https://github.com/appandflow/stim/blob/31af77d8f76bc73f1993008c45a13fd6d04895f1/packages/stim-cli/src/commands/start.ts#L461-L470) polls `isPidAlive` for its full, uninjected 5 s before the reap starts, and the test exceeds vitest's 5000 ms. `isPidAlive` counts anything but `ESRCH` as alive, and whether pid 4242 is in use on a fresh runner is nondeterministic, which is why the rerun passed.

The ENOENT is fallout: the timed-out `runAction` is still chdir'ed into its test's temp root when `afterEach` removes it, so the next helper captures a deleted directory as the cwd it restores in `finally`.

## Solution

The fake child uses `IMPOSSIBLE_PID` (above every kernel's `pid_max`), so the handoff wait returns on its first check and the SIGTERM-to-SIGKILL escalation runs on injected timing only. The `['SIGTERM', 'SIGKILL']` assertion and the reap are unchanged; this is the path the test already took whenever pid 4242 happened to be free.

`runAction` and `runConcurrentActions` restore a cwd captured once at module load, so a timed-out test cannot hand the next one a deleted directory.

**Not addressed:** an orphaned action can still fire its `console.log`/`process.exit` after the next test has installed its own mocks, which is inherent to a test that keeps running past its timeout; and the other test files use the same per-call cwd pattern, but none of their tests is known to time out, so they are left alone.

## Test plan

Linux Node 22 was not reproduced locally (no Docker daemon on this machine). Evidence instead:

- Forcing a live pid on macOS (`{ pid: process.pid }` in place of `4242`, not committed) reproduces both CI failures on the base branch: a `5011ms` timeout, then ENOENT at `start.test.ts:202` (CI: `5021ms`, same line), plus 5 knock-on failures and `EADDRINUSE` unhandled errors. With only the cwd change applied, the run drops to the forced timeout plus the next test failing with `process.exit unexpectedly called with "1"`, the pollution the ENOENT had been masking.
- `pnpm test`: 3702 passed, 2 failed in `engine-ios-device.test.ts`. Those assert the shared OS tmpdir has no `stim-devicectl-*` entries and three leftovers dated Sep 5 exist there; they fail the same way alone and are outside this diff.
- `pnpm exec vitest run packages/stim-cli/src/__tests__/start.test.ts -t "handoff write failure"`, 20 iterations while four other agents ran suites on the machine: 20/20 before (11-15 ms) and after (11-37 ms). pid 4242 is free on this Mac, so the loop cannot show the failure by itself.
- `pnpm exec vitest run packages/stim-cli/src/__tests__/start.test.ts`: 76 passed. `format:check`, `lint`, `typecheck`, `knip`, `build`: pass.

Fixes #500

### DIFF
--- a/packages/stim-cli/src/__tests__/start.test.ts
+++ b/packages/stim-cli/src/__tests__/start.test.ts
@@ -34,6 +34,7 @@ import { IMPOSSIBLE_PID, asProcessExit, makeChildProcess } from './_factories.ts
 
 let tmpHome: string;
 let root: string;
+const originalCwd = process.cwd();
 
 const successfulTunnelCleanup = async () => ({ status: 'stopped' as const });
 
@@ -164,7 +165,6 @@ async function runAction(opts: Record<string, unknown>, register: (cmd: Command)
   const origLog = console.log;
   const origErr = console.error;
   const origExit = process.exit;
-  const cwd = process.cwd();
   let exitCode = null;
   console.log = (l) => logs.push(String(l));
   console.error = (l) => errs.push(String(l));
@@ -175,7 +175,7 @@ async function runAction(opts: Record<string, unknown>, register: (cmd: Command)
   try {
     await run(opts);
   } finally {
-    process.chdir(cwd);
+    process.chdir(originalCwd);
     console.log = origLog;
     console.error = origErr;
     process.exit = origExit;
@@ -191,7 +191,6 @@ async function runConcurrentActions(opts: Record<string, unknown>, register: (cm
   const origLog = console.log;
   const origErr = console.error;
   const origExit = process.exit;
-  const cwd = process.cwd();
   console.log = (line) => logs.push(String(line));
   console.error = (line) => errs.push(String(line));
   process.exit = asProcessExit((code) => exits.push(code));
@@ -199,7 +198,7 @@ async function runConcurrentActions(opts: Record<string, unknown>, register: (cm
   try {
     await Promise.all([run(opts), run(opts)]);
   } finally {
-    process.chdir(cwd);
+    process.chdir(originalCwd);
     console.log = origLog;
     console.error = origErr;
     process.exit = origExit;
--- a/packages/stim-cli/src/__tests__/start.test.ts
+++ b/packages/stim-cli/src/__tests__/start.test.ts
@@ -1329,7 +1329,7 @@ describe('action: spawning the supervisor', () => {
           return true;
         },
       }),
-      { pid: 4242 },
+      { pid: IMPOSSIBLE_PID },
     );
     exec.spawn = (cmd, args, opts) => {
       exec.calls.spawn.push({ cmd, args, opts });


packages/stim-cli/src/__tests__/start.test.ts
```diff
@@ -1329,7 +1328,7 @@ describe('action: spawning the supervisor', () => {
           return true;
         },
       }),
-      { pid: 4242 },
+      { pid: IMPOSSIBLE_PID },
```
janicduplessis, line 1331: `IMPOSSIBLE_PID` rather than the file's usual `4242` placeholder: `waitForSupervisorHandoff` polls the real `isPidAlive(child.pid)` against a real 5 s deadline before `terminateSupervisorChild` runs, so a fake pid that exists on the host stalls the test for the whole deadline. A pid above every kernel's `pid_max` fails that check immediately, and only the injected `now`/`sleep`/`isAlive` timing is left.

janicduplessis (commented):
Verdict: CLEAR. GitHub refuses an approval from the account that authored the PR, so this is posted as a comment review.

Fresh review of 31af77d8 against `main` (aefe6ece). No blocking findings.

**Non-blocking**

1. `packages/stim-cli/src/__tests__/start.test.ts:1322-1332` — the `Object.assign(makeChildProcess({ kill }), { pid: IMPOSSIBLE_PID })` wrapper is now a no-op: `makeChildProcess` already defaults `pid` to `IMPOSSIBLE_PID` ([`_factories.ts:140`](https://github.com/appandflow/stim/blob/31af77d8f76bc73f1993008c45a13fd6d04895f1/packages/stim-cli/src/__tests__/_factories.ts#L138-L141)), and the sibling test at line 1484 builds its child with the plain `makeChildProcess({ kill(...) })` form. Optional: drop the wrapper so both SIGTERM tests construct the child the same way.

2. Regression protection is the constant plus the factory default, not a test: reintroducing a live pid would come back as the same intermittent 5 s timeout, not a deterministic failure. I do not think a deterministic guard is worth adding: it would need either a wall-clock bound on the test or a new injection seam in `waitForSupervisorHandoff` for the test's sake, and the 5 s vitest timeout is now a documented detector with #500 explaining the mechanism. Noting it so the trade-off is explicit.

3. PR description: "whether pid 4242 is in use on a fresh runner is a coin flip" states a probability nothing measures; "nondeterministic" is what the evidence supports.

Pre-existing, outside this diff: `start.test.ts:112` `DEAD_LISTENER_PID = 999999901` duplicates `IMPOSSIBLE_PID`; and the `_factories.ts:15` comment names pid_max but not the int32 ceiling: `process.kill` throws `ERR_INVALID_ARG_TYPE` for a pid >= 2^31, which `isPidAlive` would report as alive, so the constant has to stay below 2^31 as well as above pid_max (999999901 does).

**Verified**

- `waitForSupervisorHandoff` ([`start.ts:461-470`](https://github.com/appandflow/stim/blob/31af77d8f76bc73f1993008c45a13fd6d04895f1/packages/stim-cli/src/commands/start.ts#L461-L470), untouched by this diff) uses real `Date.now()`, the module-level `sleep` (`start.ts:188`) and `isPidAlive` directly; none of `registerStart`'s injected deps reach it. With a dead pid the first iteration returns `null` before any `sleep`, so the wait is removed from the test's path rather than shortened. The rest of the path is injected or immediate: `resolveSupervisorTarget` returns `none` with no record, `terminateChild` runs on the test's `now`/`sleep`/`isAlive`, and `isMetroRunning` against a closed port fails at connect.
- `process.kill(999999901, 0)` throws `ESRCH` on Node v22.12.0, v22.22.2 and v24.14.1 (macOS); `2147483648` throws `ERR_INVALID_ARG_TYPE`. 999999901 is above macOS `PID_MAX` (99999) and Linux's 2^22 `pid_max` ceiling, and below 2^31. The green Linux jobs on this SHA are Linux evidence too: a non-ESRCH result there would have reproduced the 5 s timeout.
- Base (`aefe6ece`) with `{ pid: process.pid }` in the handoff test reproduces the exact CI pair (timeout at 1315:3, `ENOENT ... chdir` at `runConcurrentActions` 202:13 from 1407:7) plus 5 knock-on failures and `EADDRINUSE` unhandled errors. Head with the same substitution: 2 failures, the forced timeout and `process.exit unexpectedly called with "1"` in the next test; the ENOENT is gone.
- Under `pool: 'forks'` each test file gets its own process, and its module-load cwd is the vitest root (probed with two throwaway test files: distinct pids, both reporting the worktree root). `originalCwd` therefore cannot be a temp dir and is captured before any hook runs regardless of test order. These two helpers are the only `chdir` sites in the file.
- The `metroExecutor` default child `{ pid: 4242 }` (line 145) does not carry the same hazard: the file passes 76/76 with that default swapped to `process.pid` and again with `IMPOSSIBLE_PID`.
- `vitest run start.test.ts`: 76 passed. Handoff test alone, 10 iterations: all pass at 11-13 ms. `pnpm test`: 3702 passed, 2 failed in `engine-ios-device.test.ts` (3 stale `stim-devicectl-*` entries in this machine's tmpdir, as the description says). `format:check`, `lint`, `typecheck`, `knip`, `build`: pass. CI on 31af77d8: node 22, node 24 and the published runtime job are green.
- No comments added; each commit is one logical change with a `test:` prefix.

janicduplessis:
Review follow-up:

- Note 3 applied: "coin flip" is now "nondeterministic" in the description.
- Note 1 not applied. The plain `makeChildProcess({ kill })` form fails `pnpm run typecheck` here because this test's `exec.spawn` must return a `ChildStub`, whose `pid` is required, while `makeChildProcess` returns a `ChildProcess` whose `pid` is optional:

  ```
  Type 'ChildProcess' is not assignable to type 'ChildStub'.
    Property 'pid' is optional in type 'ChildProcess' but required in type 'ChildStub'.
  ```

  The `Object.assign` keeps the narrowed type, so it stays.

PR #552, opened 2026-09-09
fix: distinguish pre-build refusals from missing benchmark ccache evidence

### LINKED ISSUE #494: fix: distinguish pre-build refusals from missing benchmark ccache evidence
## Problem

The Android benchmark cache audit reports ccache-evidence-missing for a platform invocation that refuses before any build, even when a later successful invocation explicitly reports an artifact cache hit. This misattributes setup failure to compiler-cache health.

## Evidence

benchmarkCcache currently visits every stim android command. A sequence containing an exit-1 STIM_NO_METRO refusal followed by an exit-0 fingerprint hit and compilation-cache-not-run message is marked investigate with ccache-evidence-missing. No private attempt results are published here.

## Cause

The platformRuns loop in scripts/agent-benchmark/run-guards.mjs requires measurements or an artifact hit from every invocation, without distinguishing a proven pre-build refusal from an attempted native build.

## Fix idea

Keep setup/launch failure evidence separate. Exempt only proven pre-build refusals from compiler-cache requirements; never exempt failed or interrupted commands that could have compiled without reporting statistics. Preserve fail-closed behavior for missing build output and retain every command in the timeline.

janicduplessis:
## Description

The Android benchmark cache audit (`benchmarkCcache` in `scripts/agent-benchmark/run-guards.mjs`) requires compiler-cache statistics or an artifact hit from every `stim android` invocation. A `stim android` that refuses with `STIM_NO_METRO` has neither, so a timeline of an exit-1 `STIM_NO_METRO` refusal followed by an exit-0 artifact hit (`compilation cache not run`) is marked `investigate` with `ccache-evidence-missing`. That blames compiler-cache health for a Metro setup failure.

## Solution

An invocation is a proven pre-build refusal when it exited 1, its output carries a listed refusal code, and nothing in the output shows a build phase (`build compiling`, a `compilation cache` line, or a `"ccache"` facts block). Only those invocations skip the per-invocation evidence requirement; the command timeline is untouched.

At least one non-refusal invocation must still exist, so a refusal that is the only platform run stays `ccache-evidence-missing`.

Only `STIM_NO_METRO` is listed, the code the report observed. Its two emission sites are inside [`resolveMetroPort`](https://github.com/appandflow/stim/blob/aefe6ecee173fdaf902b61c1f02116913ac1628d/packages/stim-cli/src/commands/android.ts#L846-L888), which `runAndroid` leaves before the fingerprint step and the [Gradle build phase](https://github.com/appandflow/stim/blob/aefe6ecee173fdaf902b61c1f02116913ac1628d/packages/stim-cli/src/commands/android.ts#L1404). Both forms [`fail` prints](https://github.com/appandflow/stim/blob/aefe6ecee173fdaf902b61c1f02116913ac1628d/packages/stim-cli/src/commands/android.ts#L697-L704) are recognized: the stderr `error  STIM_NO_METRO:` line and the `--json` `{"code":"STIM_NO_METRO",...}` payload.

`STIM_NO_PROJECT`, `STIM_WORKSPACE_STATE`, and `STIM_BAD_ARG` also fire before the build but are deliberately not listed: none was observed, and each needs its own emission-site proof.

Not handled: a wrapper that appends `echo "STATUS=$?"` records exit 0 for the shell, so a refusal run that way still fails closed, as before.

## Test plan

- `scripts/agent-benchmark/run-guards.test.mjs`: the reported sequence is `artifact-hit` with no invalid reasons, in both refusal forms; a failed build without statistics, an interrupted (`exitCode: null`), killed (137), crashed, or exit-0 command, and a lone refusal still report `ccache-evidence-missing`. The first test fails on `main` with `status: 'investigate'`; all 36 pass with the change.
- Real refusal: built the CLI and ran `stim android` and `stim android --json` in a minimal React Native fixture with no `stim start`. Both printed `error       STIM_NO_METRO: No Metro port is reserved for this workspace.`, `--json` added the `{"code":"STIM_NO_METRO",...}` payload, exit 1. `benchmarkCcache` on that captured output plus an exit-0 artifact hit returns `artifact-hit`; on the refusal alone, `investigate` with `ccache-evidence-missing`.
- `pnpm run format:check`, `pnpm run lint`, `pnpm run build`, `pnpm run typecheck`, `pnpm run knip`: pass.
- `pnpm test` (with `NODE_USE_SYSTEM_CA` unset): 3705 passed, 3 failed, all pre-existing and unrelated. `gc.test.ts` timed out under load and passes alone; two `engine-ios-device.test.ts` tests assert `tmpdir()` holds no `stim-devicectl-*` entries and this machine has leftovers.

Fixes #494

### DIFF
--- a/scripts/agent-benchmark/README.md
+++ b/scripts/agent-benchmark/README.md
@@ -62,9 +62,9 @@ Collection records actual hits and misses, accepts proven artifact hits without
 C++ compilation, and flags missing or below-target compiler-cache evidence. A
 `stim android` invocation that exits 1 with a `STIM_NO_METRO` refusal and no
 build phase output stays in the timeline but does not count as missing
-compiler-cache evidence; another invocation must still produce build output, and
-a failed, interrupted, or crashed command that could have compiled still fails
-closed.
+compiler-cache evidence; another `stim android` invocation must still carry
+compiler statistics or a proven artifact hit, and a failed, interrupted, or
+crashed command that could have compiled still fails closed.
 Completed tool output triggers an immediate `CACHE ALERT` and preserves
 `cache-alerts.json`. A flagged attempt stays available for investigation and is
 excluded from published comparisons; investigate the cause before retrying.
--- a/scripts/agent-benchmark/README.md
+++ b/scripts/agent-benchmark/README.md
@@ -59,7 +59,12 @@ Android native Stim cells also require `ccacheMinHitRatePercent`. Establish this
 machine/scenario threshold from a verified warm compiler-cache probe before
 dispatch; the example value is illustrative. Keep it fixed across models.
 Collection records actual hits and misses, accepts proven artifact hits without
-C++ compilation, and flags missing or below-target compiler-cache evidence.
+C++ compilation, and flags missing or below-target compiler-cache evidence. A
+`stim android` invocation that exits 1 with a `STIM_NO_METRO` refusal and no
+build phase output stays in the timeline but does not count as missing
+compiler-cache evidence; another invocation must still produce build output, and
+a failed, interrupted, or crashed command that could have compiled still fails
+closed.
 Completed tool output triggers an immediate `CACHE ALERT` and preserves
 `cache-alerts.json`. A flagged attempt stays available for investigation and is
 excluded from published comparisons; investigate the cause before retrying.
--- a/scripts/agent-benchmark/run-guards.mjs
+++ b/scripts/agent-benchmark/run-guards.mjs
@@ -409,15 +409,27 @@ function artifactCacheHit(entry) {
   );
 }
 
+const preBuildRefusalCodes = ['STIM_NO_METRO'];
+
+function preBuildRefusal(entry) {
+  return (
+    entry.exitCode === 1 &&
+    !/build\s+compiling|compilation cache|"ccache"\s*:/.test(entry.output) &&
+    preBuildRefusalCodes.some((code) =>
+      new RegExp(`"code"\\s*:\\s*"${code}"|^\\s*error\\s+${code}:`, 'm').test(entry.output),
+    )
+  );
+}
+
 export function benchmarkCcache(meta, commands) {
   const minimum = meta.timingTarget?.ccacheMinHitRatePercent ?? null;
   const result = { minimumHitRatePercent: minimum, status: 'not-applicable', builds: [], invalidReasons: [] };
   if (meta.arm !== 'stim' || meta.platform !== 'android') return result;
   if (meta.variant === 'native' && minimum == null) result.invalidReasons.push('ccache-target-missing');
-  const platformRuns = commands.filter(
-    (entry) => commandSegmentsStartingWith(entry.command, 'stim android').length > 0,
+  const attemptedBuilds = commands.filter(
+    (entry) => commandSegmentsStartingWith(entry.command, 'stim android').length > 0 && !preBuildRefusal(entry),
   );
-  for (const entry of platformRuns) {
+  for (const entry of attemptedBuilds) {
     const measurements = ccacheMeasurements(entry.output);
     result.builds.push(...measurements.map((measurement) => ({ commandId: entry.id ?? null, ...measurement })));
     if (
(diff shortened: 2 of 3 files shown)


janicduplessis (commented):
Fresh review of the diff against `main`. No correctness findings. Two non-blocking notes, then what was verified.

## Notes (non-blocking)

1. `scripts/agent-benchmark/README.md:65` — "another invocation must still produce build output" does not match the rule the code enforces. `benchmarkCcache` (`scripts/agent-benchmark/run-guards.mjs:432-446, 463`) requires at least one non-refusal `stim android` invocation, and each such invocation must carry compiler statistics or a proven artifact hit. An artifact hit produces no build output, and the sentence before this one already says artifact hits are accepted. Suggested wording: "another `stim android` invocation must still carry compiler statistics or a proven artifact hit".

2. `scripts/agent-benchmark/run-guards.mjs:412-420` — `preBuildRefusalCodes` is a one-element list with a per-code dynamic `RegExp`, while the PR body says the other pre-build codes are deliberately not listed. A single literal regex (`/"code"\s*:\s*"STIM_NO_METRO"|^\s*error\s+STIM_NO_METRO:/m`) expresses the same rule with less machinery (CLAUDE.md "Simplicity first"). Optional.

## Verified

- Emission sites: the string `STIM_NO_METRO` is defined once for Android (`packages/stim-cli/src/commands/android/support.ts:99`, `NO_METRO`) and used in `android.ts` only at 852 and 869, both inside `resolveMetroPort` (846), called at 888. Everything before 888 is settings and flag validation, `readProductFlavors` (parses `build.gradle` text, `engine/gradle.ts:388-390`), `systemImageRefusal`, and project registration; nothing compiles. `expo prebuild -p android --no-install` (`engine/prebuild.ts:124`) runs at 1364 and Gradle at 1404-1405 after `phase('build', 'compiling ...')`, both after the gate. The other emitters (`ios.ts:608/629`, `reload.ts:263/274`) reach a `stim android` timeline entry only through a compound command; there, a compiled android segment prints `build compiling` on stderr (phase lines print in `--json` mode too), and a successful segment prints the `compilation cache` facts line (`commands/android/result.ts:269`) or a `"ccache":` JSON fact, so the exemption does not apply.
- Real CLI (invariant 9): built the branch and ran `stim android` and `stim android --json` in a copy of `__tests__/fixtures/sample-bare-project` with `STIM_HOME` redirected and no `stim start`. Output (non-TTY, no ANSI) was exactly `  error       STIM_NO_METRO: No Metro port is reserved for this workspace.` plus the remedy line, exit 1; `--json` added `{"code":"STIM_NO_METRO","message":...,"remedy":...}`. Feeding the captured output to `benchmarkCcache`: refusal + artifact hit -> `artifact-hit` with no reasons in both forms; refusal alone -> `investigate` / `ccache-evidence-missing`; the same refusal text with exit 0 or `null` -> `ccache-evidence-missing`; `stim android; echo "STATUS=$?"` (shell exit 0) -> `ccache-evidence-missing`, as the PR body states.
- Exit-code recording: `scripts/agent-benchmark/command-evidence.mjs:105-109` maps a Claude-runner result without an integer exit code to 1 on `is_error`/`interrupted`, so exit 1 alone does not prove a refusal. The text gate carries the proof, and the code text is only emitted before the build, so an interrupted or killed compile is not admitted.
- No other consumer changed: `platformRuns` in `benchmarkSetupInvalidReasons` (301) and `platformCommands` in `benchmarkTiming` (330) are separate locals; `stale-cmake-launcher-state` still scans `commands`; `driver.mjs:2165` passes `commandAudit.commands` unfiltered and only reads `ccache.invalidReasons` and the `ccache` record, so every command stays in the timeline.
- Tests assert `status` and `invalidReasons` from `benchmarkCcache`; none checks a constant or source text. Swapping in `origin/main`'s `run-guards.mjs` reproduces the claimed failure of the first new test (`status: 'investigate'`); the other 35 pass on both.
- PR body matches the diff and the permalinks (`fail` 697/704, `resolveMetroPort` 846-888, build phase 1404). No comments added; no iteration narration.
- Commands: `vitest run scripts/agent-benchmark/run-guards.test.mjs` (36 passed); `vitest run scripts/agent-benchmark` (80 passed, 1 skipped); `pnpm run lint`, `format:check`, `typecheck`, `build`, `knip` pass; `gh pr checks 552`: test (node 22), test (node 24), published runtime (node 20.19.4) pass, native suites skipped.

## Considered and dismissed

- A compound command whose earlier segment prints the `stim guide facts` example (`{ "code": "STIM_NO_METRO", ... }`, `guide/facts.ts:248`) followed by `stim android --json 2>/dev/null` that compiled and failed with exit 1 would be exempted. It needs stderr discarded and a second segment supplying the code text; not a realistic benchmark command, noted only.
- `stim doctor ...; stim android` refusal is not exempted because doctor output contains "compilation cache": a fail-closed miss, not a false positive.
- An ANSI-colored `error` line would not match the refusal regex: fail-closed, consistent with the other audit regexes, and real non-TTY output is uncolored.

janicduplessis:
Review follow-up:

- Note 1 applied: the README now says another `stim android` invocation must carry compiler statistics or a proven artifact hit, matching what `benchmarkCcache` enforces.
- Note 2 left as is: the explicit code list is deliberate, so adding a second pre-build code is a one-line change with its own emission-site proof rather than a regex edit.

PR #554, opened 2026-09-09
fix: bundle nested dynamic imports in the bare Metro server

### LINKED ISSUE #476: bare-inproc Metro fails to bundle a file with a path-aliased dynamic import()
## Problem

`stim android` builds, installs, and launches the app successfully, but Metro
(running in Stim's own `bare-inproc` supervisor) fails to bundle a file that
contains a dynamic `await import(...)` call, with:

```
Unable to resolve module ./node_modules/metro-runtime/src/modules/asyncRequire.js from src/example/handleExample.ts:

None of these files exist:
  * src/example/node_modules/metro-runtime/src/modules/asyncRequire.js(...)
  * src/example/node_modules/metro-runtime/src/modules/asyncRequire.js
```

Expected: the same bundle that a plain `react-native start` Metro produces
without error (see Evidence) — i.e. the dynamic import resolves and the app's
initial bundle serves cleanly.

## Evidence

- Stim `1.0.0-rc.19`, Node `v22.22.0`, macOS 26.5.1 (arm64)
- Project: bare React Native 0.86 app, `metro-runtime@0.84.4`
- Reproduction:
  ```bash
  ENVFILE=.env.staging stim start
  ENVFILE=.env.staging stim android --variant stagingDebug --json
  ```
  Result JSON (identifying fields replaced with placeholders):
  ```json
  {"platform":"android","serial":"emulator-5556","avdName":"stim-example","deviceName":"stim-example","systemImage":"system-images;android-36;google_apis;arm64-v8a","fingerprint":"aefd638e28b17404540fe34dc5c0e3ca1fb40239","cacheKey":"aefd638e28b17404540fe34dc5c0e3ca1fb40239-stagingdebug-sim-arm64-v8a","variant":"stagingDebug","metroPort":8082,"cacheHit":false,"cacheSkipped":false,"waitedForBuild":null,"appPath":"/path/to/android/app/build/outputs/apk/staging/debug/app-staging-debug.apk","bundleId":"com.example.app.staging","installSkipped":false,"launched":"bundling","ccache":{"status":"reported","hits":10,"misses":569,"hitRatePercent":1.7},"debugHttpHost":"10.0.2.2:8082","debugHttpHostNote":null,"devClientUrl":null,"logs":"/path/to/.stim/workspaces/example-project-hash/logs","durationMs":304673}
  ```
  Then:
  ```bash
  stim logs --errors --json
  ```
  surfaces the bundling error quoted above (`event":"bundling_error"`, `buildID":"1"`, `platform":"android"`).

- **Not reproducible outside Stim.** Same repo, same file, same `ENVFILE`,
  plain Metro instead of Stim's supervisor:
  ```bash
  ENVFILE=.env.staging npx react-native start --port 8199 --reset-cache
  curl "http://localhost:8199/index.bundle?platform=android&dev=true&minify=false" -o bundle.js
  ```
  returns HTTP 200, a complete ~52MB bundle, and no errors in the Metro log.

- Sanity check that the module itself is resolvable from the project root:
  ```bash
  node -e "console.log(require.resolve('metro-runtime/src/modules/asyncRequire'))"
  # -> /path/to/project/node_modules/metro-runtime/src/modules/asyncRequire.js (absolute, correct)
  ```
  `metro-runtime`'s `package.json` `exports` map also correctly exposes
  `./modules/asyncRequire` -> `./src/modules/asyncRequire.js`.

- The failing source line is a dynamic import reached via a path alias (not a
  relative import), e.g. `await import('@features/example/useExampleHook')`,
  inside a `switch`/conditional branch of the affected file.

## Cause

Unknown. What's ruled out:
- Not a missing/misconfigured dependency — `metro-runtime` resolves fine from
  Node and from a plain Metro process in the same repo.
- Not caused by the project's custom `resolver.resolveRequest` in
  `metro.config.js` — that hook only rewrites a few specific module names and
  falls through to `context.resolveRequest` for everything else, including
  this import.

Best guess, unverified: Stim's `bare-inproc` Metro supervisor builds or passes
the merged Metro config (specifically `transformer.asyncRequireModulePath`,
which `@react-native/metro-config`'s `getDefaultConfig` normally sets to an
absolute path via `require.resolve`) differently than the project's own
`metro.config.js` entry point does when Metro is started the normal way — e.g.
if config construction or transform-worker option passing happens in a
different process/cwd context inside Stim, the value could end up unset and
fall back to Metro core's own relative-path default instead of react-native's
absolute one. This is a guess from the symptom, not confirmed by reading
Stim's source.

## Fix idea

No fix proposed — filing this to report the discrepancy between Stim's
in-process Metro and a standalone `react-native start` Metro on the same
project/config. Whoever picks this up will want to compare how
`bare-inproc` mode constructs/forwards `transformer.asyncRequireModulePath`
(and possibly other transformer options) versus the plain CLI path, ideally
with a minimal repro project containing one path-aliased dynamic `import()`.
Closing this needs a `stim android`/`stim ios` run on an affected project
bundling that file without the error, not just a plausible-looking diff.

janicduplessis:
## Description

`stim start` on a bare React Native project cannot bundle a file below the project root that contains a dynamic `import()`. Metro answers the bundle request with HTTP 500:

```
Unable to resolve module ./node_modules/metro-runtime/src/modules/asyncRequire.js from src/example/handleExample.ts
```

Stim rewrites `transformer.asyncRequireModulePath` to the project-relative `./node_modules/...` string ([bc28b54e](https://github.com/appandflow/stim/commit/bc28b54e18d4b03a01f250ce1dad4ff69198bb33), for a worktree-independent transform key). Metro emits that string verbatim as a dependency of every module with an `import()` ([collectDependencies.js#L276](https://github.com/react/metro/blob/4ac5f345c9964db6110eda9250f1a657d0f11cfd/packages/metro/src/ModuleGraph/worker/collectDependencies.js#L276-L277)) and resolves it relative to the importing file ([resolve.js#L468](https://github.com/react/metro/blob/4ac5f345c9964db6110eda9250f1a657d0f11cfd/packages/metro-resolver/src/resolve.js#L468)), so only a root-level file finds it. Plain `react-native start` keeps the absolute path and is unaffected.

## Solution

Rewrite the path to Metro's own default, the bare specifier `metro-runtime/src/modules/asyncRequire` ([defaults/index.js#L93](https://github.com/react/metro/blob/4ac5f345c9964db6110eda9250f1a657d0f11cfd/packages/metro-config/src/defaults/index.js#L93)): node_modules lookup resolves it from any directory, and the transform key stays root-independent because it hashes the config value ([metro-transform-worker/src/index.js#L778](https://github.com/react/metro/blob/4ac5f345c9964db6110eda9250f1a657d0f11cfd/packages/metro-transform-worker/src/index.js#L778)).

**Scope:** the rewrite only fires for the hoisted `<root>/node_modules/metro-runtime/src/modules/asyncRequire.js`. A pnpm `.pnpm/...` path or a custom module stays absolute: it bundles, but keeps a per-worktree transform key as before bc28b54e.

**Alternative:** resolving the relative form against the project root in a `resolveRequest` wrapper would intercept every resolution, including the project's own resolver, for the same result.

## Test plan

Real Metro runs (invariant 9) on a fresh `@react-native-community/cli init` app (RN 0.87.1, metro 0.87.0, npm) where `src/example/handleExample.ts` does `await import('@features/example/useExampleHook')` through a `resolveRequest` path alias plus a nested relative `import()`; `STIM_HOME` in a scratch directory, CLI built from this branch:

- Unfixed (aefe6ece): `stim start --json`, then `curl "http://localhost:8082/index.bundle?platform=android&dev=true&minify=false"` returns 500 with the error above; `stim logs --errors --json` shows `bundling_error`.
- This branch: the same request returns 200 (4.1 MB) with `metro-runtime/src/modules/asyncRequire`, `src/example/handleExample` and `src/features/example/useExampleHook` in the bundle; `stim logs --errors --json` prints nothing; `stim stop` frees the port.
- Cross-worktree key: Metro's `getTransformCacheKey` on two copies of the app at different paths differs with the absolute path (`e6348d74...` vs `d5cfbe4d...`) and matches with the rewrite (`33783d9e...`).
- Not run: `stim android` on the fixture (no cached build or emulator for it), and a pnpm-installed project.

`pnpm run format:check`, `lint`, `build`, `typecheck`, `knip` and `test:e2e` pass.

**Known failing:** `pnpm test` has two pre-existing `engine-ios-device.test.ts` failures that expect no `stim-devicectl-*` directories in the machine tmpdir; three stale ones from Sep 5 sit there, unrelated to this change.

Fixes #476

### DIFF
--- a/packages/stim-cli/src/__tests__/supervisor-bare.test.ts
+++ b/packages/stim-cli/src/__tests__/supervisor-bare.test.ts
@@ -171,14 +171,35 @@ describe('config adjustments', () => {
     expect(ensureNativePlatform({})).toBe(false);
   });
 
-  test('normalizes a project-local Metro transformer path across worktrees', () => {
+  test('rewrites the metro-runtime asyncRequire path to the specifier Metro resolves from any module (#476)', () => {
     const config = {
       transformer: {
         asyncRequireModulePath: join(root, 'node_modules', 'metro-runtime', 'src', 'modules', 'asyncRequire.js'),
       },
     };
     expect(normalizeMetroTransformerPaths(config, root)).toBe(true);
-    expect(config.transformer.asyncRequireModulePath).toBe('./node_modules/metro-runtime/src/modules/asyncRequire.js');
+    expect(config.transformer.asyncRequireModulePath).toBe('metro-runtime/src/modules/asyncRequire');
+  });
+
+  test('keeps an asyncRequire path that is not the hoisted metro-runtime module', () => {
+    for (const path of [
+      join(
+        root,
+        'node_modules',
+        '.pnpm',
+        'metro-runtime@0.84.5',
+        'node_modules',
+        'metro-runtime',
+        'src',
+        'modules',
+        'asyncRequire.js',
+      ),
+      join(root, 'src', 'asyncRequire.js'),
+    ]) {
+      const config = { transformer: { asyncRequireModulePath: path } };
+      expect(normalizeMetroTransformerPaths(config, root)).toBe(false);
+      expect(config.transformer.asyncRequireModulePath).toBe(path);
+    }
   });
 
   test('keeps a Metro transformer path outside the project root', () => {
--- a/packages/stim-cli/src/supervisor/server-bare.ts
+++ b/packages/stim-cli/src/supervisor/server-bare.ts
@@ -180,12 +180,18 @@ function installSharedCacheStore({
   return true;
 }
 
+// Metro's collectDependencies emits this value verbatim as a dependency of every
+// module with an import(), so a project-relative path only resolves from the project root.
+// Metro's default bare specifier resolves from any directory, but only to the copy
+// hoisted to node_modules.
+const ASYNC_REQUIRE_MODULE = 'metro-runtime/src/modules/asyncRequire';
+
 export function normalizeMetroTransformerPaths(config: BareModule, root: string): boolean {
   const current = config?.transformer?.asyncRequireModulePath;
   if (typeof current !== 'string' || !isAbsolute(current)) return false;
-  const local = relative(root, current);
-  if (local === '' || local === '..' || local.startsWith(`..${sep}`) || isAbsolute(local)) return false;
-  config.transformer.asyncRequireModulePath = `./${local.split(sep).join('/')}`;
+  const local = relative(root, current).split(sep).join('/');
+  if (local !== `node_modules/${ASYNC_REQUIRE_MODULE}.js`) return false;
+  config.transformer.asyncRequireModulePath = ASYNC_REQUIRE_MODULE;
   return true;
 }
 


janicduplessis (commented):
**Verdict: CLEAR.** No correctness defect found. Two wording items below are minor and non-blocking.

### Findings

1. PR description, first paragraph: "cannot bundle a file **outside** the project root that contains a dynamic `import()`". The failing file is inside the project, in a subdirectory of the root (`src/example/handleExample.ts` in the repro); only files directly in the root directory could resolve the old `./node_modules/...` specifier. "Outside the project root" describes a different situation. Suggest "a file below the project root" or "in a subdirectory of the project root".

2. [server-bare.ts#L183-L186](https://github.com/appandflow/stim/blob/c1d29fa3c82d9d0a905a800f7a1cea89351acd0a/packages/stim-cli/src/supervisor/server-bare.ts#L183-L186): the comment says `metro-transform-worker` emits the value verbatim. metro-transform-worker only forwards `config.asyncRequireModulePath` as an option (`metro-transform-worker/src/index.js:224` in 0.87.0); the string literal is built and emitted by `metro/src/ModuleGraph/worker/collectDependencies.js`, the file the description links. The comment is kept under the "external dependency constraint" exception, which asks for the source to be named, so name Metro's `collectDependencies`.

### Verified

- Root cause: `@react-native/metro-config` 0.86.0 and 0.87.1 set `asyncRequireModulePath` with `require.resolve(...)` (absolute); metro-config 0.80.12, 0.84.4 and 0.87.0 default to the bare `metro-runtime/src/modules/asyncRequire`; `collectDependencies` emits the configured string verbatim; metro-resolver joins a `./` specifier onto the importing module's directory. The four Metro permalinks in the description point at the cited lines at `4ac5f345`.
- Exact-path condition: `root` is `realpathSync`'d in [supervisor/run.ts#L122](https://github.com/appandflow/stim/blob/c1d29fa3c82d9d0a905a800f7a1cea89351acd0a/packages/stim-cli/src/supervisor/run.ts#L122) and the config value comes from Node's `require.resolve`, so both sides are canonical; `relative()` absorbs trailing separators; `split(sep).join('/')` covers Windows. A pnpm `.pnpm/...` realpath and a custom `src/asyncRequire.js` stay untouched (unit test and the real run below).
- Package exports: metro-runtime 0.81.5 through 0.87.0 export `./src/*` -> `./src/*.js`, so the bare specifier resolves through the exports map under `unstable_enablePackageExports: true` (default since metro-config 0.82) with no "Falling back" warning; 0.80.12 has no exports map and package exports default off, so file-based resolution applies. Real run: 0 of 1305 log records contain "Falling back".
- `disableHierarchicalLookup` / `nodeModulesPaths`: the bare specifier resolves from every directory that can resolve `react-native` itself; a config that disables hierarchical lookup without listing the app's `node_modules` breaks every bare import, so no working project loses a resolution the absolute path gave it. Against the previous `./` form, which failed for every nested file, this is strictly better.
- Transform key: `getCacheKey` hashes `remainingConfig`, which includes `asyncRequireModulePath`, so rc.19 cache entries carrying the `./` dependency string are not reused. Re-ran `keycheck.cjs` on two clonefile copies of the fixture: absolute path -> `e6348d74...` vs `62cb99b2...`; rewritten -> `33783d9e...` on both.
- Tests: both asyncRequire tests fail against `origin/main`'s `server-bare.ts` (`expected './node_modules/metro-runtime/src/modu...' to be 'metro-runtime/src/modules/asyncRequire'`, `expected true to be false`) and pass on the branch; a reintroduced `./` form fails the first assertion. No e2e exercises a nested `import()` (the bare fixture is a fresh init app and `launched: 'bundling'` is tolerated), so the Metro-side behavior rests on manual runs. Not blocking; a follow-up could add a nested `import()` module to the e2e fixture.
- Real Metro (invariant 9), CLI built from this branch in a separate worktree, `STIM_HOME` in a scratch directory, no device created or booted:
  - npm fixture (RN 0.87.1 / metro 0.87.0, `src/example/handleExample.ts` with an aliased and a relative `import()`): `stim start`, then `GET /index.bundle?platform=android&dev=true&minify=false` -> 200, 4,112,000 bytes; `metro-runtime/src/modules/asyncRequire` x3, `src/features/example/useExampleHook` x1, "Unable to resolve module" x0; `stim logs --errors --json` 0 records; the `config_adjusted` rewrite record is present; `stim stop` freed 8082.
  - pnpm fixture (`node-linker=isolated` + `shamefully-hoist=true`; `node_modules/metro-runtime` is a symlink and the config value realpaths to `.pnpm/metro-runtime@0.87.0/...`): no rewrite record, bundle 200, 4,205,903 bytes, same module counts, 0 error records. A plain pnpm-isolated install fails earlier on the pre-existing `STIM_BARE_DEPS` check (metro not resolvable from the root), unrelated to this change.
  - `stim android` not run (no emulator, and no device creation from the review). The #476 error is Metro's bundling error and independent of the device, so the bundle request exercises the failing operation.
- Checks: `vitest run supervisor-bare.test.ts` 39/39; `pnpm run lint`, `typecheck`, `format:check`, `knip` clean; `pnpm test` 3703 passed, 2 failed in `engine-ios-device.test.ts` (stale `stim-devicectl-*` tmpdirs on this machine, as the description states); changed files ASCII-only; `gh pr checks 554`: test (node 22), test (node 24), published runtime pass, ios/android matrix skipped. Branch is 3 commits behind `main` with no overlap in the touched files.
- Docs: no guide topic or website page mentions `asyncRequireModulePath` or the transform-key normalization; the debug log message stays accurate.

janicduplessis:
Review follow-up: both notes applied. The description now says "below the project root", and the comment above `ASYNC_REQUIRE_MODULE` names Metro's `collectDependencies` as the emitter (amended into the commit, no other change).

PR #555, opened 2026-09-09
test: stub react-native config in the fingerprint fixture

### LINKED ISSUE #491: test: avoid implicit React Native downloads in fingerprint fixture

janicduplessis:
## Description

`scripts/agent-benchmark/cache-key.test.mjs` downloads `react-native` from the npm registry on every cold run, which is what pushed it past its 60 s timeout in #491. The pinned `@expo/fingerprint` (0.20.10) collects React Native core autolinking sources for a project without `expo-modules-autolinking` by spawning `npx react-native config` in the project root ([Bare.ts](https://github.com/expo/expo/blob/52fc013ec8c75e89ba8f83c167e4b6b0416e6bf7/packages/@expo/fingerprint/src/sourcer/Bare.ts#L94)). The fixture has no `react-native`, so npx installs it into its cache ([libnpmexec](https://github.com/npm/cli/blob/58c302d73123cd5a841568d17e896d52aacc8755/workspaces/libnpmexec/lib/index.js#L148-L160)), and even a warm cache revalidates the manifest against the registry on every call.

## Solution

Give the fixture an executable `node_modules/.bin/react-native` that prints an empty autolinking config. npx checks the local bin directory before it considers installing anything, so it runs the stub and never contacts the registry, while the pinned fingerprint code still runs end to end, including the guard against loading the fixture's conflicting `@expo/fingerprint` package. The timeout, the fingerprint options, and `cache-key.mjs` are unchanged. The stub also writes a marker file that the test asserts after the first fingerprint, so a silent fallback to the registry fails the test instead of only slowing it down.

`npm_config_yes=false` for this test was the other option; it stops the install, but npx still resolves the manifest from the registry first, so the test would keep making network requests. The stub is a `/bin/sh` script, which covers the ubuntu and macos CI runners and the platforms Stim supports.

## Test plan

`env -u NODE_USE_SYSTEM_CA pnpm exec vitest run scripts/agent-benchmark/cache-key.test.mjs` with an empty npm cache (`npm_config_cache=<empty dir>`, `npm_config_offline` unset):

- `origin/main`: 4 passed in 9.82 s on a fast network. npm logs `warn exec The following package was not found and will be installed: react-native@0.87.1`, fetches `react-native` and its dependency manifests from `registry.npmjs.org`, and the cache gains 163 MB under `_npx`.
- This branch: 4 passed in 0.83 s. The cache ends with no `_npx` directory and no `http fetch` line in the npm logs, and a `ps` poll during the run shows `npm exec react-native config` executing the fixture's `node_modules/.bin/react-native` stub.

**`pnpm test`**: 3701 passed, 1 skipped, 3 failed, none related to this diff: two in `engine-ios-device.test.ts` assert the OS tmpdir has no `stim-devicectl-*` entries and three stale ones exist on this machine, and one `gc.test.ts` timeout under load that passes alone (142/142). `format:check`, `lint`, `build`, `typecheck`, and `knip` pass.

Fixes #491

### DIFF
--- a/scripts/agent-benchmark/cache-key.test.mjs
+++ b/scripts/agent-benchmark/cache-key.test.mjs
@@ -1,5 +1,5 @@
 import { describe, expect, it } from 'vitest';
-import { mkdtempSync, mkdirSync, writeFileSync, rmSync } from 'node:fs';
+import { existsSync, mkdtempSync, mkdirSync, writeFileSync, rmSync } from 'node:fs';
 import { tmpdir } from 'node:os';
 import { join } from 'node:path';
 import { fileURLToPath } from 'node:url';
@@ -23,9 +23,17 @@ it('hashes native inputs with the pinned CLI dependency despite a conflicting fi
       join(root, 'node_modules/@expo/fingerprint/index.js'),
       'throw new Error("Fixture fingerprint must not be loaded");',
     );
+    // @expo/fingerprint runs `npx react-native config`; without this local bin, npx installs react-native from the registry.
+    mkdirSync(join(root, 'node_modules/.bin'));
+    writeFileSync(
+      join(root, 'node_modules/.bin/react-native'),
+      '#!/bin/sh\n: > "$PWD/react-native-config.ran"\nprintf \'{"root":"%s","dependencies":{}}\' "$PWD"\n',
+      { mode: 0o755 },
+    );
     writeFileSync(join(root, 'android/build.gradle'), 'version = "one"');
     writeFileSync(join(root, 'ios/native.m'), 'one');
     const initial = benchmarkFingerprint(root, stimPackage, 'android');
+    expect(existsSync(join(root, 'react-native-config.ran'))).toBe(true);
     expect(initial).toMatch(/^[a-f0-9]{40}$/);
     writeFileSync(join(root, 'android/local.properties'), 'sdk.dir=/machine-specific-sdk');
     writeFileSync(join(root, 'ios/native.m'), 'two');


janicduplessis (commented):
**Verdict: CLEAR.** No defects found in the diff. One optional suggestion.

### Optional

[`scripts/agent-benchmark/cache-key.test.mjs:26-32`](https://github.com/appandflow/stim/blob/82f011255784cd20e3dc8d82a2ec68036c4add7a/scripts/agent-benchmark/cache-key.test.mjs#L26-L32): nothing asserts the stub was used. `@expo/fingerprint` swallows any failure of `npx react-native config` and returns `[]` ([Bare.ts#L105-L108](https://github.com/expo/expo/blob/52fc013ec8c75e89ba8f83c167e4b6b0416e6bf7/packages/@expo/fingerprint/src/sourcer/Bare.ts#L105-L108)), and a registry install still finishes under the 60 s timeout on a fast network (latest `main` CI run, 67a6da2c: this file took 14360 ms on node 22 and 14529 ms on node 24; the #491 symptom needs a slow network or Node 26). If npx ever stops resolving the fixture's local bin, the test silently returns to the pre-fix behavior. A deterministic check is a marker written by the stub (for example `: > "$PWD/react-native-config.ran"`) asserted after the first `benchmarkFingerprint` call; a file at the fixture root is outside the hashed `android` dir, so the hash assertions are unaffected. Counter-argument: that asserts npm's bin resolution rather than Stim behavior, so skipping it is defensible.

### Verified

- `@expo/fingerprint` 0.20.10 (lockfile pin) spawns `npx react-native config` with `cwd: projectRoot` and `getOriginalEnv()` ([Bare.ts#L94](https://github.com/expo/expo/blob/52fc013ec8c75e89ba8f83c167e4b6b0416e6bf7/packages/@expo/fingerprint/src/sourcer/Bare.ts#L94)); the fixture has no `expo-modules-autolinking`, so `Sourcer.ts` resolves `useRNCoreAutolinkingFromExpo` to false and takes this path. `parseCoreAutolinkingSourcesAsync` ([Bare.ts#L179-L198](https://github.com/expo/expo/blob/52fc013ec8c75e89ba8f83c167e4b6b0416e6bf7/packages/@expo/fingerprint/src/sourcer/Bare.ts#L179-L198)) iterates `config.dependencies` and uses `root` only inside that loop. Dumping `sources` from `createFingerprintAsync` on an identical fixture gives `dir android` (`bareNativeDir`) plus `contents rncoreAutolinkingConfig` = `{}`: the stub's JSON is consumed, `android/build.gradle` still drives the hash, and `ios/native.m` stays excluded for platform `android`, which is what the stability assertion relies on.
- npm 10.9.7 / libnpmexec 9.0.4 (installed here, byte-identical to the permalinked SHA): `hasPkgBin` finds no `bin` in the fixture `package.json`, then `localFileExists` walks up from `node_modules/.bin`, finds the stub, and calls `run()` ([index.js#L148-L154](https://github.com/npm/cli/blob/58c302d73123cd5a841568d17e896d52aacc8755/workspaces/libnpmexec/lib/index.js#L148-L154)) before any Arborist load, manifest fetch, or install. The `npm_config_yes=false` claim holds: `getManifest` uses `preferOnline: true` ([L25-L31](https://github.com/npm/cli/blob/58c302d73123cd5a841568d17e896d52aacc8755/workspaces/libnpmexec/lib/index.js#L25-L31)) and runs at [L178-L193](https://github.com/npm/cli/blob/58c302d73123cd5a841568d17e896d52aacc8755/workspaces/libnpmexec/lib/index.js#L178-L193), before the `yes === false` throw at [L266-L273](https://github.com/npm/cli/blob/58c302d73123cd5a841568d17e896d52aacc8755/workspaces/libnpmexec/lib/index.js#L266-L273).
- Reproduction with a fresh `npm_config_cache` (forced; `~/.npm/_npx` here already held 187 MB): `origin/main` 9.75 s, `_npx` 163 MB, 455 `http fetch` lines and `warn exec ... react-native@0.87.1` in the first npm log, calls 2 and 3 revalidate (`cache revalidated`). This branch: 0.92 s, no `_npx`, 0 `http fetch` lines across all three npm logs, `verbose cwd` = fixture root.
- Ubuntu (`/bin/sh` is dash): this PR's CI jobs report `cache-key.test.mjs` at 2879 ms (node 22) and 5847 ms (node 24) against 14360 ms and 14529 ms on the latest `main` run.
- `$PWD` is a `printf` argument, not part of the format string, so `%` in the path is inert. The shell resets `PWD` to the real cwd when the inherited value does not match (observed: the stub saw `/private/var/.../benchmark fingerprint X` while the parent's `PWD` was the repo root). A `"` or `\` in `tmpdir()` would break the JSON, but that does not occur on the CI runners or macOS, and `root` does not reach the hash: a `{"dependencies":{}}`-only stub gives the same hash and sources because `maybeGetRealPathAsync` catches the rejection. No Windows runner exists (`ci.yml` ubuntu, `e2e-native.yml` macos), and `website/docs/requirements.md` lists macOS and Linux hosts.
- Hygiene: the stub is written with mode `0o755` before the first fingerprint call; `rmSync(root, ...)` in `finally` removes it; nothing lands in the project tree. The comment names the external dependency and the concrete constraint, so it fits the comment policy.
- PR description: every claim matched the code or my runs; no iteration-history narration; template sections present. Bare.ts#L94 at 52fc013 is the spawn line and that SHA's package.json is 0.20.10; libnpmexec#L148-L160 at 58c302d is the local-bin check and fallthrough, and that SHA is npm 10.9.7 / libnpmexec 9.0.4.
- Checks: `pnpm exec vitest run scripts/agent-benchmark/cache-key.test.mjs` 4 passed in 0.9 s; `lint`, `format:check`, `knip` exit 0; `gh pr checks 555` green (test node 22/24, published runtime), native e2e skipped as expected. `main` has 3 commits past the merge base, none touching this area.

janicduplessis:
Review follow-up: the optional note is applied. The stub now writes `react-native-config.ran` in the fixture root and the test asserts it after the first `benchmarkFingerprint` call, so a fallback to the registry fails deterministically instead of only taking longer. The marker lives outside the hashed `android` directory, and the stability and change assertions are unchanged. Amended into the commit.

PR #557, opened 2026-09-09
fix: require guide loading in benchmark prompts

### LINKED ISSUE #556: Make the benchmark guide prerequisite explicit
## Problem

The launch-error benchmark can successfully diagnose and repair an app with Stim but still fail publication with `stim-guide-agent-missing-or-failed`. Repeated Sonnet runs follow the explicit worktree/warm/start/launch instructions without running the required guide command.

## Evidence

`scripts/agent-benchmark/driver.mjs` builds the Stim task as `git worktree add`, then `stim worktree warm`. The supplied Stim skill separately requires `stim guide agent` before using Stim. The audit rejects completed runs that omit it, even when the initial platform command captures the injected error and Settings proof succeeds.

## Cause

The generated task spells out the operational sequence but omits a prerequisite required by the skill and validator. This does not prove why a model ignores the skill, but makes the task unnecessarily easy to follow incompletely.

## Fix idea

Include `stim guide agent` explicitly before `stim worktree warm` in the generated Stim task for both platforms. Preserve the existing guide audit, timing, control instructions, and raw failed attempts. Verify the emitted prompt order and a fresh run that actually reads the guide.

janicduplessis:
## Description

Sonnet launch-error runs can complete the repair and Settings proof but fail the audit because they skip `stim guide agent`. The task explicitly lists warm/start/launch, while the guide prerequisite appears only in the supplied skill.

## Solution

Put the existing guide prerequisite before warm-up in the task itself. Keep control instructions, timing, and validation unchanged; failed attempts remain private rather than being relabeled valid.

## Test plan

The launch-crash prompt self-test verifies guide-before-warm ordering on iOS and Android and no Stim guide instruction on control. Run a fresh Stim attempt and confirm its command transcript reads the guide before warm-up.

Fixes #556

### DIFF
--- a/scripts/agent-benchmark/driver.mjs
+++ b/scripts/agent-benchmark/driver.mjs
@@ -909,7 +909,7 @@ function promptFor(arm, variant, runId, runDir, crash = null, requestedPlatform 
   const stimSource = variant === launchCrashVariant ? crash.fixtureCheckout : main;
   const worktree =
     arm === 'stim'
-      ? `In ${stimSource}, run exactly \`git worktree add -b worktree-bench/${runId} ${quotedStimPath} HEAD\`. Then change into ${stimPath}, run \`stim worktree warm\`, and wait for its successful exit before running start or a platform command. Work only in that checkout. `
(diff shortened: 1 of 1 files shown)


janicduplessis:
Independent code review is clear at cc63dbdd95. The fresh Sonnet iOS transcript now successfully runs `stim guide agent` before `stim worktree warm`; the full run continues without intervention. All required CI checks pass. Local full units have only the already-tracked #491 fingerprint timeout; 3,703 pass, and 66 end-to-end checks plus runtime and website checks pass.

PR #561, opened 2026-09-09
fix: keep benchmark endpoint circles visible

### LINKED ISSUE #560: Keep benchmark endpoint circles fully visible
## Problem

The first Agent circle is cut in half at time zero on the benchmark timeline. Endpoint event circles can also lose their right half at the horizontal scroller edge.

## Evidence

User screenshot shows the first purple circle clipped against the sticky Agent label on the mobile benchmark details page. Current main has the same timeline styles.

## Cause

`BenchmarkTimeline.module.css` centers circles on their timestamp with `translate(-50%, -50%)`, but the time tracks have no inset. At zero the left half sits behind the opaque sticky label; at the end the scroller can clip the right half.

## Fix idea

Give the axis and every timeline track the same horizontal gutter, keeping timestamps aligned while reserving room for endpoint circles and their halos. Verify both endpoints, mobile scrolling, and zoom.

janicduplessis:
## Description

The time-zero circle sits half behind the sticky Agent label. The same overlap affects event markers.

## Solution

Inset the axis and all tracks equally, leaving room for endpoint circles and their halos without moving timestamps relative to commands. Shell and background rows shrink with the inset instead of overflowing their time scale. This is CSS-only; benchmark data and timing are unchanged.

## Test plan

Compared old and new styles in 390px browser frames at 1x and 2x zoom, scrolled to both ends. The first circles are fully visible after the fix, with commands and axis positions aligned; the final markers and halos also fit.

Before and after at time zero, 2x zoom

Fixes #560

### DIFF
--- a/website/src/components/BenchmarkTimeline.module.css
+++ b/website/src/components/BenchmarkTimeline.module.css
@@ -251,9 +251,16 @@
   padding: 0.5rem 0;
 }
 
+.axis,
+.dotTrack,
+.shellTracks,
+.backgroundTracks {
+  margin-inline: 1rem;
+}
+
 .backgroundTrack {
   position: relative;
-  min-width: 44rem;
+  min-width: 0;
 }
 
 .backgroundBar {
@@ -389,7 +396,7 @@
 
 .shellTrack {
   position: relative;
-  min-width: 44rem;
+  min-width: 0;
 }
 
 .commandBar {


janicduplessis:
Independent CSS review is clear. All local checks passed, including 3,737 unit tests and website build/typecheck; required exact-head CI is green. Replaced the accidentally cropped visual evidence with a verified full before/after screenshot; attachment URLs are hosted by GitHub.

janicduplessis:
Also verified the production-built Opus iOS benchmark page: the complete time-zero dot and final dot are visible, aligned with the command tracks.

stim560-actual

PR #572, opened 2026-09-09
perf: copy warm worktrees directly

### LINKED ISSUE #571: Speed up worktree warm by avoiding per-file publication
## Problem

`stim worktree warm` takes 79-86 seconds in the latest Trailhead startup-crash benchmarks, despite successful same-volume APFS cloning. Warming is now a substantial part of the task.

## Evidence

On macOS 26.5 / Node 26.7.0, Trailhead fixture 45c134e86833523737da87df17c55bb9fa7a2547, external APFS SSD: current copier took 79.840s in a phase profile (25s cloning, 40s publication, 15s staging cleanup). An interleaved fresh baseline took 69.490s; three direct `cp -Rc` trials took 22.010s, 22.615s, and 22.126s. Direct copying with up-front destination/ancestor checks took 22.550s; those checks cost 0.307ms combined. These are isolated copy experiments, not full build benchmarks.

## Cause

`cloneIgnoredEntries` in `packages/stim-cli/src/worktree.ts` clones ignored trees into private staging, then republishes every file and deletes staging. This preserves concurrent-writer protection introduced in #420; #427 retained APFS sharing during publication, and #434 kept staging on the operation volume. The extra passes still dominate metadata-heavy trees.

## Fix idea

Use direct CoW copies into missing destinations, keeping the cheap tracked/existing-path/ancestor guards and exclusions. Eliminate intermediate staging and preserve fallback behavior. Document that concurrent writes to the destination are unsupported: warm must finish before installs, builds, Metro, or other writers begin. The user explicitly accepts this trade-off. Keep source isolation and avoid internal-volume staging. Validate exclusions, fallback/partial failures, read-only modes, symlinks, and real APFS performance before merging.

janicduplessis:
## Description

Warming Trailhead's ignored files took 69s despite same-volume APFS cloning. The copier now takes 27s with destination checks and exclusions retained. These are copy timings, not full mobile-build benchmarks; most of the removed time was per-file publication and staging cleanup.

## Solution

Copy directly into missing destinations. Keep the up-front tracked-file, existing-entry, symlink-ancestor, and nested-worktree checks; remove excluded generated output from the copy. Doctor checks the actual source/destination volumes instead of a staging path warm no longer uses.

This deliberately relaxes [the concurrent-writer protection introduced in #420](https://github.com/appandflow/stim/commit/64121692796dacd0b24e90fe2115346f49d42a46). **Other processes must not write to the destination until warm finishes:** late files can be overwritten or removed. The agent guide and user docs state this constraint. Copy failures remain nonzero and can leave partial output to inspect before retrying.

## Test plan

- Real Git/filesystem regressions cover existing entries, source isolation, symlinks, read-only directories, generated-cache exclusions, and partial-copy fallback.
- A disposable APFS-volume test exercises direct copying and forced clone fallback with system temp and `STIM_TMPDIR` on another volume, then checks doctor's cross-volume finding for an actual cross-volume worktree.
- Trailhead fixture `45c134e86833523737da87df17c55bb9fa7a2547`, external APFS SSD, macOS 26.5 / Node 26.7.0: 27.06s and 27.05s. All nine entries copied without fallback; sampled files matched and had independent inodes.

Fixes #571.

### DIFF
--- a/packages/stim-cli/README.md
+++ b/packages/stim-cli/README.md
@@ -93,6 +93,10 @@ there. Warm copies missing ignored state from main, including eligible `.env`
 and local configuration files. Existing entries are preserved; existing
 ignored directories such as `node_modules` are skipped whole.
 
+Wait for warm to finish before editing, installing dependencies, starting
+Metro/builds, or running another warm in that worktree. Concurrent writes are
+unsafe: entries created after the initial check can be overwritten or removed.
+
 After the work is preserved, `stim worktree remove` removes any linked
 worktree, warmed or not. Git-created branches stay. See the
 [worktree guide](https://appandflow.github.io/stim/docs/worktrees) for exclusions
--- a/packages/stim-cli/src/__tests__/doctor-storage.test.ts
+++ b/packages/stim-cli/src/__tests__/doctor-storage.test.ts
@@ -50,7 +50,7 @@ test('doctor detects iOS output/cache and configured staging mismatches separate
   expect(findings[1]?.detail).toContain(output);
 });
 
-test('an explicit override on another volume warns for warm and artifact preparation', () => {
+test('an explicit override on another volume warns for artifact preparation but not direct warming', () => {
   const staging = join(base, 'override');
   const findings = checkStorageLayout(project, {
     platform: 'ios',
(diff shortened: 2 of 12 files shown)


janicduplessis:
Independent code review of dd48459b8054a42c2d634ce4f0ea07990a7a7997 is clear: no actionable findings. The reviewer checked direct-copy fallback, exclusions, symlinks, source independence, permissions, storage diagnostics, and the documented concurrent-write trade-off, and independently passed 61 focused tests. Required CI is green on Node 22/24 and the Node 20.19.4 published-runtime check. Production-copy trials were 27.06s and 27.05s; this PR remains draft for user review.

PR #577, opened 2026-09-09
fix: preserve shared-build waits after native preparation

### LINKED ISSUE #576: Preserve shared-build results across prebuild cache-key changes
## Problem

Two cold Expo Android worktrees racing the same initial fingerprint compile only once, but the waiter reports a failed build and returns `waitedForBuild: null` after successfully installing the builder's artifact under the post-prebuild key.

The output says `the build lock was released without an artifact` and `FAILED without an artifact`, although the builder completed successfully. The cache QA single-flight row consequently fails.

## Evidence

Reproduced on current main,49ae2fce16fbc49b69c710063ded7b685f3c7fea, using:

```sh
node test/e2e/native/run-cache-e2e.mjs --framework expo --platform android --summary /tmp/expo-android-cache.json
```

The first fingerprint changes after Expo prebuild. The builder compiles once and stores the post-mutation artifact. The waiter waits for the old-key lock, receives the false failure warning, performs its own prebuild, then hits the same post-mutation cache key without compiling. Both apps launch successfully. Both result payloads have `waitedForBuild: null`; the check expects exactly one waiter.

## Cause

`engine/build-lock.ts` `waitForBuild` resolves only the original key and treats a released lock without an artifact at that key as builder failure. The Android command populates `waitedForBuild` only when that wait directly returns an artifact. It loses the successful wait when artifact resolution occurs later, after this workspace's own native preparation changes its key.

## Fix idea

Preserve accurate wait evidence across native preparation and report an unknown old-key outcome without asserting the build failed. Recheck this workspace's own post-mutation fingerprint before using any artifact. Keep storing artifacts only under the post-mutation key; do not copy artifacts under old keys or trust another workspace's changed fingerprint without validating this workspace. Check the analogous iOS path and retain real failed-builder diagnostics. Rerun the cold Expo Android single-flight case with unchanged assertions.

janicduplessis:
## Description

When Expo prebuild changes the fingerprint during a shared build, the waiter looks for the artifact under the old key. It incorrectly reports that the builder failed, then finds the actual artifact after its own prebuild but returns `waitedForBuild: null`. The Android cache suite reproduces this even though only one workspace compiled and both apps launched.

Fixes #576.

## Solution

Treat a released lock as an unknown old-key outcome, not proof of failure. On iOS and Android, a waiter that later finds a matching artifact retains its wait metadata and prints the normal cache-handoff result. It still prepares and fingerprints its own workspace before accepting the artifact; storage remains exclusively under the final key.

The dead-builder takeover warning from [d31e702f25](https://github.com/appandflow/stim/commit/d31e702f25093ff72d5335c3b64cfa697656d32e) stays; a released lock also allows a fresh build, but does not claim failure.

## Test plan

The real Expo Android cache suite passed on bbd7956: six checks passed, two iOS-only checks skipped. The unchanged race assertions observed one compiler and a cache-hit waiter retaining a 41.1s wait, without the false failure warning. Both race launches verified bundle loading, and all five cleanup checks passed.

Run the cold Expo Android cache suite:

```sh
node test/e2e/native/run-cache-e2e.mjs --framework expo --platform android --summary /tmp/cache-summary.json
```

Metro reused all 714 transforms; Gradle restored 89 tasks from cache. The auxiliary forced-compile and same-tree reinstall checks returned launch unverified (no fresh bundle request); those prove caching, not a new healthy launch. Initial cold and cross-worktree cached launches were verified. Regression coverage also checks the equivalent iOS post-preparation hit and rebuilding when no artifact exists.

### DIFF
--- a/packages/stim-cli/src/__tests__/android-command.test.ts
+++ b/packages/stim-cli/src/__tests__/android-command.test.ts
@@ -2130,18 +2130,22 @@ describe('single-flight builds', () => {
     expect(h.stdout.length).toBe(1);
   });
 
-  test('a builder that failed makes the waiter take over and build', async () => {
+  test.each([false, true])('a missing artifact is rebuilt after a released lock: %s', async (released) => {
     let acquires = 0;
     const h = harness({
       acquireLock: () => (++acquires === 1 ? heldBy() : { acquired: true, path: '/lock', lock: { pid: process.pid } }),
-      waitForBuild: async () => ({ builderFailed: 'the build lock was released without an artifact', waitedMs: 4000 }),
+      waitForBuild: async () =>
+        released
+          ? { lockReleased: true as const, waitedMs: 4000 }
+          : { builderFailed: 'the builder (pid 41233) is gone', waitedMs: 4000 },
     });
     const result = await h.run();
     expect(result.ok).toBe(true);
     expect(acquires).toBe(2);
     expect(h.calls.build.length).toBe(1);
     expect(h.calls.releaseLock.length).toBe(1);
-    expect(h.stderr.join('\n')).toMatch(/without an artifact/);
+    expect(result.facts?.waitedForBuild).toBeNull();
+    expect(/FAILED without an artifact|RETRY:/.test(h.stderr.join('\n'))).toBe(!released);
   });
 
   test('losing the takeover race waits for the new holder and installs its artifact', async () => {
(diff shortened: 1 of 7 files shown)


janicduplessis:
Fresh independent code review is clear at bbd7956f04d68ad261756e309bfea9b3cc38f6eb: cache safety, released/dead lock handling, fallbacks, regression tests and guide reviewed; ten targeted cases passed. Real Expo Android cache QA has now passed with the unchanged race assertions and complete owned cleanup. Exact-head Node 22, Node 24 and published-runtime CI are green.

PR #579, opened 2026-09-09
fix: allow busy iOS simulators more time to launch

### LINKED ISSUE #578: iOS simulator launch hits a 10-second timeout during concurrent cold boots
## Problem

The bare iOS cache-concurrency suite intermittently refuses an installed app with `STIM_LAUNCH_FAILED: simctl launch ... failed: spawnSync xcrun ETIMEDOUT`. Three normal runs reproduced the failure after concurrent owned-simulator boots; compilation and artifact reuse succeeded.

## Evidence

Run `node test/e2e/native/run-cache-e2e.mjs --framework bare --platform ios` on current main (`fd65de16ecdd5f970d95aaa3d14ad387feba4182`). In the instrumented reproduction, the original launch call timed out after 10,009 ms with empty stdout/stderr. The other race participant launched in 680 ms. The build ran once, with 115/115 compilation-cache hits and a 44.2-second cache waiter. An immediate read-only process probe found no matching app process after the failed launch.

A separate diagnostic allowing 60 seconds passed, but every launch in that run took under one second: it did not reproduce the slow path and is not evidence that a longer deadline fixes the underlying simulator delay.

## Cause

`launchIosApp` passes a hard-coded 10-second timeout to both direct and initial-URL simulator launches. That deadline was added with crash capture in `dfd342114e6edc926f1d74ef26adc985560744ae`; it remains on current main. The reason CoreSimulator stalls in some concurrent boots is not established.

## Fix idea

Allow a longer, still-bounded simulator launch deadline without retrying the launch or relaxing readiness/crash checks. Validate the real concurrent-launch workflow and retain timeouts as failures. A truly hung launch would take longer to refuse; this is not a claim that simulator contention is solved.

janicduplessis:
## Description

Concurrent cold simulator boots can leave `stim ios` failing at its 10-second `simctl launch` deadline after a successful cached build. Three concurrent-boot QA runs hit this timeout; the underlying delay remains unexplained.

## Solution

Increase the [deadline added with crash capture](https://github.com/appandflow/stim/commit/dfd342114e6edc926f1d74ef26adc985560744ae) to 60 seconds for direct and initial-URL simulator launches. No launch retries or readiness/crash checks change. A genuinely hung launch takes up to 50 seconds longer to refuse.

## Test plan

The real branch rerun passed all seven applicable cache checks and owned cleanup. It exercised the slow path: one concurrent `simctl launch` succeeded after **53.112 seconds**, while the other took 762 ms. Both apps passed launch verification. Exactly one build compiled (115/115 compilation-cache hits); the other workspace waited 46.155 seconds and installed that artifact.

Reproduce with `node test/e2e/native/run-cache-e2e.mjs --framework bare --platform ios`. This demonstrates a launch that the old deadline would abort; it does not explain or eliminate CoreSimulator's intermittent delay.

Fixes #578.

### DIFF
--- a/packages/stim-cli/src/engine/app-install.ts
+++ b/packages/stim-cli/src/engine/app-install.ts
@@ -271,7 +271,7 @@ export function launchIosApp(
           // project directly; launch-then-openurl can create two React hosts.
           const initialUrl = new URL(url).searchParams.get('url')!;
           const pid = parseLaunchedPid(
-            e.runFile('xcrun', [...launchArgs, '--initialUrl', initialUrl], { timeoutMs: 10000 }),
+            e.runFile('xcrun', [...launchArgs, '--initialUrl', initialUrl], { timeoutMs: 60000 }),
           );
           return { ok: true, mode: 'launch', url, jsLocation: jsLocationValue(metroPort), pid };
         }
@@ -284,7 +284,7 @@ export function launchIosApp(
   }
 
   try {
-    const out = e.runFile('xcrun', launchArgs, { timeoutMs: 10000 });
+    const out = e.runFile('xcrun', launchArgs, { timeoutMs: 60000 });
     const result: IosLaunchResult = { ok: true, mode: 'launch', pid: parseLaunchedPid(out) };
     if (metroPort !== null) result.jsLocation = jsLocationValue(metroPort);
     return result;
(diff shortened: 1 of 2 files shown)


janicduplessis:
Fresh code review is clear at 0967a533382edfafccb39cf4d051423cf3f4c9e7. The real iOS cache suite now passes (7 checks, 1 Android-only skip), including a measured 53.112-second successful simctl launch, both verified racing apps, single compilation, and all owned cleanup checks. All three required exact-head CI jobs passed.

PR #585, opened 2026-09-09
fix: recognize already populated Metro QA caches

### LINKED ISSUE #583: Avoid labeling settled nonempty Metro QA caches as EMPTY
## Problem

The native cache QA suite calls a populated, unchanged Metro transform store `EMPTY` and waits 90 seconds for new growth even when the bundle has already finished and populated that store before observation starts.

This is a test-helper reporting/delay issue, not evidence of broken Metro caching. The suite's separate before/after counts correctly prove storage and reuse.

## Evidence

Observed in both iOS and Android real-project cache QA with candidate 822116234268aa96a3d89cc15b6bf26675f5014b (the subsequent main change d17e4fd30c does not alter this helper).

Android printed:

```text
wt1 Metro store: no growth at all in 90s (2838 files, unchanged) -- reported as EMPTY, not as settled
```

The subsequent check measured 0 -> 2838 entries for the first worktree and 2838 -> 2839 for the second, with both loaded Expo configurations confirming the shared store. `metro-store` passed with 100% rounded reuse. iOS showed the same message with 2754 existing entries.

## Cause

`settle()` in `test/e2e/native/run-cache-e2e.mjs` takes its own baseline after the build, initializes `moved` to false, and requires additional observed growth before the quiet-period path can finish. A nonempty store that is already quiet takes the first-growth timeout and is labeled empty. The helper does return the real counts; the wording does not match them.

## Fix idea

Distinguish an empty store awaiting its first write from a nonempty store awaiting stability. Keep the suite's independent pre-build growth and cross-worktree reuse assertions, configuration confirmation, and bounded observation. Test already-populated/quiet, initially-empty/delayed-write, and continuously-changing cases. Do not silently turn a truly empty store into a passing cache check.

janicduplessis:
## Description

The native cache QA suite waited 90 seconds and labeled a populated Metro store `EMPTY` when it had finished filling before observation began. Fixes #583.

## Solution

Let nonempty stores settle after the existing six-second quiet period. Keep the first-write wait for empty stores: the [original guard](https://github.com/appandflow/stim/blob/0a8b1e3786d036b62067a291afbc364fd8af0408/test/e2e/native/run-cache-e2e.mjs#L961-L985) prevents a delayed cold write from being mistaken for a settled cache. Cache-growth/reuse assertions and the overall timeout are unchanged; this affects QA only, not published CLI behavior.

## Test plan

- Exercised the exact helper with real temporary directories and a controlled clock: prefilled, empty, delayed writes, continuous writes, byte-size-only changes, and zero-byte entries. The old helper reproduces the 90-second/`EMPTY` result; the fix preserves the empty/delayed-write and timeout behavior.
- Measured the helper against a completed real iOS QA Metro store: 2,757 entries settled in 6.3 seconds using the real clock.

### DIFF
--- a/test/e2e/native/run-cache-e2e.mjs
+++ b/test/e2e/native/run-cache-e2e.mjs
@@ -900,22 +900,20 @@ async function settle(dir, label, { firstGrowthMs = 90000, quietMs = 6000, timeo
   const startedAtMs = Date.now();
   const baseline = dirStats(dir);
   let last = baseline;
-  let moved = false;
+  let populated = baseline.files > 0;
   let stableSince = Date.now();
   while (Date.now() - startedAtMs < timeoutMs) {
     await sleep(stepMs);
     const now = dirStats(dir);
     if (now.files !== last.files || now.bytes !== last.bytes) {
       last = now;
-      moved = true;
+      populated ||= now.files > 0;
       stableSince = Date.now();
       continue;
     }
-    if (!moved) {
+    if (!populated) {
       if (Date.now() - startedAtMs >= firstGrowthMs) {
-        log(
-          `${label}: no growth at all in ${Math.round(firstGrowthMs / 1000)}s (${baseline.files} files, unchanged) -- reported as EMPTY, not as settled`,
-        );
+        log(`${label}: still empty after waiting ${Math.round(firstGrowthMs / 1000)}s for cache entries`);
         return last;
       }
       continue;


janicduplessis:
Fresh independent code review is clear at bc03d5b3b5fb08d4cbc98ba861f9172756ac9176. The reviewer independently reran the populated/empty/delayed-write/continuous-write/byte-change/zero-byte probes and measured the completed native QA Metro store settling in 6.2 seconds. These targeted probes are private verification evidence, not committed regression tests. Required CI is green on the reviewed commit.

PR #590, opened 2026-09-10
fix: resolve app schemes in renamed iOS workspaces

### LINKED ISSUE #589: Resolve shared app schemes when the workspace name differs
## Problem

`stim ios` refuses a bare React Native example whose shared app scheme is visible to Xcode, but does not have the workspace's name. It then incorrectly tells the user to share the already-shared scheme.

## Evidence

Found during stable-release field QA using react-native-safe-area-context at `63fab204dd0ba980c8d867d3ef9ccf2c1fd76f13` (React Native 0.85.3, react-native-test-app 5.3.2). The example has workspace `RNSACExample`, `app.json` name `safe-area-example`, and shared schemes `safe-area-example` and `ReactTestApp`, plus many CocoaPods schemes. After the repository's normal dependency and Pod preparation, `xcodebuild -list -json` lists both app schemes.

The candidate at `822116234268aa96a3d89cc15b6bf26675f5014b` refuses with `STIM_NO_SCHEME`: “No buildable scheme found” and “Share the app scheme in Xcode”. Current main `120d4ab93186e98e0600cacc51bd8870a56f842e` has the same selection code. No compilation starts; the QA-owned simulator was stopped and retained for cleanup.

## Cause

`pickScheme` in `packages/stim-cli/src/engine/xcode.ts` accepts a workspace-name match or a single non-test scheme. It treats multiple listed schemes as ambiguous, even when the app's declared name identifies an exact listed scheme. `resolveScheme` reports ambiguity as if no shared scheme existed.

## Fix idea

Preserve workspace-name priority and refusal for unresolved ambiguity. When the workspace name does not match, use the static React Native `app.json` name if Xcode actually lists that scheme; do not run project scripts or guess from scheme ordering. Distinguish an ambiguous listing from an empty one in the remedy. Validate against the real example and cover the fallback, priority, malformed/missing app name, and still-ambiguous cases.

janicduplessis:
## Description

`stim ios` rejected the safe-area-context example even though Xcode listed its shared app scheme: the workspace is named `RNSACExample`, while the app scheme is `safe-area-example`, alongside many CocoaPods schemes. It then suggested sharing an already-shared scheme. Fixes #589.

## Solution

Keep workspace-name and sole non-test scheme selection unchanged. If those cannot choose, match the static top-level `app.json` name against Xcode's actual scheme list. Missing, malformed, or unmatched names still refuse ambiguous selection; no project config is executed and no flags are added. The refusal and detailed guides now explain how to resolve a naming mismatch instead of treating every failure as an unshared scheme.

## Test plan

- Extended the real-Xcode test: rename a working workspace and add another shared scheme; selection fails until `app.json` identifies the intended scheme.
- Exercise populated Pod-style scheme lists, missing/malformed/non-string/unlisted app names, workspace-name priority, and the separate empty-list remedy.
- The actual prepared safe-area-context example now resolves `safe-area-example` from Xcode's 87 listed schemes without changing its workspace or scheme files.

### DIFF
--- a/packages/stim-cli/src/__tests__/engine-xcode.test.ts
+++ b/packages/stim-cli/src/__tests__/engine-xcode.test.ts
@@ -1,7 +1,7 @@
 import assert from 'node:assert';
 import type { ChildProcess, SpawnOptions } from 'node:child_process';
 import { EventEmitter } from 'node:events';
-import { existsSync, mkdirSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from 'node:fs';
+import { existsSync, mkdirSync, mkdtempSync, readFileSync, renameSync, rmSync, writeFileSync } from 'node:fs';
 import { tmpdir } from 'node:os';
 import { join } from 'node:path';
 import { getExecutor, resetExecutor, setExecutor } from '../exec.ts';
@@ -268,7 +268,7 @@ describe('parseSchemeList', () => {
 });
 
 describe('pickScheme', () => {
-  test('the scheme named after the container wins, which is every generated RN project', () => {
+  test('the scheme named after the container wins', () => {
     expect(pickScheme(['MyApp', 'MyApp-tvOS', 'MyAppTests'], 'MyApp')).toBe('MyApp');
   });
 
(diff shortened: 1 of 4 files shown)


janicduplessis:
Fresh independent code review is clear at 7fe3f957367d65745803c50da4e74e0f9fd64418. The unchanged real safe-area-context example now completes the full iOS command: 2m 21s total, 1m 54s native compilation, launched: true, and zero matching errors from logs --errors. The tracked fixture remains clean. This verifies the actual shared scheme is usable, not just present in the listing. Required CI is green on this commit.

PR #610, opened 2026-09-10
fix: relaunch stuck Android apps with an explicit launcher intent

### LINKED ISSUE #603: Use a deterministic Android relaunch remedy without monkey key failures
## Problem

The Android unverified-launch remedy suggests `am force-stop` followed by `monkey -p <app> 1`. On an emulator without physical keys, the monkey command can exit 251 with `SYS_KEYS has no physical keys but with factor 2.0%`, leaving recovery incomplete. An explicit app relaunch through the existing agent-device session works.

## Evidence

- Observed in 1.0.0 QA, candidate cc0cb528c8ecdf54619c393ee0fc084f05c2f374, bare React Native 0.85.3 sample, Android 36 Google APIs Play Store PS16K arm64 emulator.
- After restoring an intentionally missing import, the cached platform launch returned `unverified` because no new bundle request arrived. The exact printed force-stop/monkey remedy exited 251 with the key warning above.
- Existing-session `agent-device open <app> --platform android --serial <owned-serial> --metro-port <workspace-port> --relaunch` then succeeded. Stim recorded a fresh completed bundle, the repaired sample was visible, and `logs --errors --json` was empty.
- This is recovered QA friction, not a claim that the unverified platform invocation itself proved a healthy launch.

## Cause

The printed remedy relies on Android monkey's random-event setup even though the requested action is just a deterministic app launch. The exact key-configuration failure is observed; whether monkey launched an activity before returning that exit code is not established.

## Fix idea

Prefer a deterministic app-scoped launch remedy that works without physical keys, preserving the exact owned serial, package and Metro target. Verify it against a no-physical-keys emulator and retain the requirement to inspect fresh bundle/UI evidence. Do not change launch verification semantics or hide unverified results.

janicduplessis:
## Description

When `stim android` cannot verify a launch, the recovery text tells the agent to restart the app with `adb shell am force-stop <pkg> && adb shell monkey -p <pkg> 1`. On emulator images with no physical keys, `monkey` exits 251 with `SYS_KEYS has no physical keys but with factor 2.0%`, so the printed remedy stalls before the app restarts.

## Solution

The remedy now prints `am start -n <component>` with the launcher activity Stim already resolved for its own launch through `cmd package resolve-activity`. The launch result carries that component into the unverified-launch lines.

Printed deep-link commands preserve quoting through both the host shell and adb's device shell, so URL query parameters and the trailing intent extras survive copy/paste.

An implicit `am start -a MAIN -c LAUNCHER -p <pkg>` is not a substitute: activity starts add `MATCH_DEFAULT_ONLY`, and the React Native template's `MainActivity` filter has no `CATEGORY_DEFAULT`, so it fails with "unable to resolve Intent" on a real app. A dev-client launch that went through a deep link never resolved a component, so its remedy is the force-stop followed by the same deep-link command, kept as one line an agent can paste. Only when neither exists does the `monkey` fallback remain, which matches the launch path's own fallback in `launchAndroidApp`.

## Test plan

- On a booted Pixel_9 AVD (API 36) with the trailhead Expo dev client installed, `adb shell am start -n com.appandflow.trailhead/.MainActivity` prints `Starting: Intent { cmp=com.appandflow.trailhead/.MainActivity }` and exits 0. The implicit `am start -a android.intent.action.MAIN -c android.intent.category.LAUNCHER -p com.appandflow.trailhead` fails with `Error: Activity not started, unable to resolve Intent` and exit 1 on the same app. This AVD declares physical keys, so the exit 251 from the issue did not reproduce here; the new command never consults that configuration.
- Follow-up on Stim-owned `emulator-5556`, Android 36 arm64, Expo 58 canary / RN 0.87: executed the exact rendered resend, deep-link restart and explicit-component restart commands through `/bin/sh` and real adb. Android activity state retained the complete URL including `&disableFab=1`, and `am start` retained the extras. Both force-stop paths created new app PIDs (6578 → 6801 → 6848); the app rendered Trails and `stim logs --errors --json` was empty. The shell regression additionally covers an apostrophe in the URL.

Fixes #603

### DIFF
--- a/packages/stim-cli/src/__tests__/android-command.test.ts
+++ b/packages/stim-cli/src/__tests__/android-command.test.ts
@@ -2804,7 +2804,9 @@ describe('launch verification', () => {
     const text = h.stderr.join('\n');
     expect(text).toMatch(/UNVERIFIED/);
     expect(text).toMatch(/DEVELOPMENT SERVERS/);
-    expect(text).toMatch(/adb -s emulator-5584 shell monkey -p com\.example\.app 1/);
+    expect(text).toContain(
+      'adb -s emulator-5584 shell am force-stop com.example.app && adb -s emulator-5584 shell am start -n com.example.app/.MainActivity',
+    );
     expect(text).not.toMatch(/simctl/);
     expect(h.stdout.join('\n')).toMatch(/UNVERIFIED/);
     expect(h.stdout[0]).toContain(phaseLine('metro', 'state unverified on port 8082'));
--- a/packages/stim-cli/src/__tests__/engine-app-install.test.ts
+++ b/packages/stim-cli/src/__tests__/engine-app-install.test.ts
@@ -1102,18 +1102,36 @@ describe('unverifiedLaunchLines', () => {
     expect(text).toMatch(/xcrun simctl launch --console U1 com\.x/);
   });
 
-  test('Android recovery restarts the reported app before launching it again', () => {
+  test('Android recovery restarts the reported app through its resolved launcher activity', () => {
     const text = unverifiedLaunchLines({
       platform: 'android',
       metroPort: 8082,
       bundleId: 'com.x',
       serial: 'emulator-5584',
+      component: 'com.x/.MainActivity',
     }).join('\n');
     expect(text).not.toMatch(/simctl/);
     expect(text).toContain(
+      'adb -s emulator-5584 shell am force-stop com.x && adb -s emulator-5584 shell am start -n com.x/.MainActivity',
+    );
+    expect(text).not.toMatch(/monkey/);
+    expect(text).toMatch(/DEVELOPMENT SERVERS/);
+  });
+
+  test('a dev-client launch without a resolved activity restarts through the deep link, not monkey', () => {
+    const base = { platform: 'android', metroPort: 8082, bundleId: 'com.x', serial: 'emulator-5584' };
+    const url = devClientUrl('com.x', 8082);
+    const deepLink = unverifiedLaunchLines({ ...base, devClientUrl: url }).join('\n');
+    expect(deepLink).toContain(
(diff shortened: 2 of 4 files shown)


janicduplessis:
Coordinating overnight with the Claude implementation owner per Janic: see PR613 coordination comment. I will not duplicate this fix. Please confirm current head has fresh review, real adb launcher-intent verification, and is ready for merge; Codex will integrate reviewed/green fixes into the 1.0.0 release candidate and repeat affected QA.

janicduplessis:
Fresh Codex review found one P2 at app-install.ts:1315 (head4578d2d14): the printed deep-link restart only quotes for the host shell. adb joins shell argv without escaping, so the generated &disableFab=1 URL is split by the device shell; am receives a truncated URL and --ez is executed separately (exit127). Reviewer reproduced the exact rendered command through two shells; deviceShellArg(deviceShellArg(url)) preserved URL and extras. Existing resend line1307 shares it. Please fix both rendered lines and add a two-shell final-argv regression, then post fresh review/real validation readiness. Holding merge; no implementation edits by Codex. Component propagation and fallback ordering otherwise look correct.

janicduplessis:
The release integration is now waiting on the two-shell URL quoting correction noted above. If your implementation session is still active, please push that fix and post readiness; otherwise Codex can take only this review follow-up in a separate checkout, preserve your original commit, and use a normal fast-forward push. No source edits to this branch yet. PR615 integration is underway separately; native workloads remain coordinated.

janicduplessis:
Taking the narrow review follow-up now in a separate checkout so the release can proceed: quote both rendered deep-link commands through the host and device shells, add a final-argv regression, retain your launcher-activity behavior, then repeat fresh review and exact-head CI. I will only fast-forward the branch; any concurrent remote commits will be preserved. No unrelated refactor or new feature.

PR #614, opened 2026-09-10
feat: select an explicit iOS build scheme

### LINKED ISSUE #605: Allow explicit iOS Xcode scheme selection
## Problem

Stim cannot explicitly select an iOS Xcode scheme. Projects with several app
schemes must currently match the workspace/project or static app.json name,
even when the caller knows which app or environment to build. The user approved
adding `stim ios --scheme <name>`, analogous to Android's variant selector.

## Evidence

Current main cc0cb528c8ecdf54619c393ee0fc084f05c2f374 exposes iOS configuration,
device and runtime selectors but no scheme selector. `resolveScheme` in
`packages/stim-cli/src/engine/xcode.ts` only performs automatic selection.
Commit 5aa719fcfd85b56bdd19b954d6e644ed20c17e56 improves that fallback but cannot
express an explicit caller choice. Both Expo and React Native CLI support an
explicit Xcode scheme.

## Cause

The iOS command does not carry a selected scheme through the build and artifact
cache paths. Exposing only the existing engine argument would be insufficient:
two explicit schemes with the same native fingerprint must not share an artifact
or a single-flight build lock accidentally.

## Fix idea

Add a validated, noninteractive `ios --scheme <name>` while preserving automatic
selection when omitted. Keep the selected scheme consistent in build arguments,
artifact selection, cache keys, shared-build locks and provider behavior. Refuse
unknown names with available choices instead of editing the project's scheme or
app.json. Document the option and test custom/missing schemes, cache separation,
unchanged automatic selection and a real Xcode invocation. This flag selects an
Xcode scheme, not the dev client's URL scheme.

janicduplessis:
## Description

Some apps have several shared Xcode schemes, and automatic selection cannot express which one to build. `stim ios --scheme "App Staging"` selects an exact name while preserving automatic selection when omitted.

Fixes #605.

## Solution

An explicit scheme gets separate artifact keys, shared-build locks and DerivedData so another scheme's app cannot be reused. Exact, case-sensitive names stay isolated. Stim validates existing schemes before cache lookup and refuses ambiguous products after a build rather than installing another app.

The legacy Expo project-provider tier is skipped for explicit schemes because its artifacts may not be scheme-keyed. Stim's keyed local and configured cache providers remain available. This build selector does not change the dev client's URL scheme.

## Test plan

- Real Xcode simulator build: shared scheme `Staging App` produces `Scratch.app`, with the correct bundle ID and scheme-specific DerivedData.

No physical-device signing or remote-device session was exercised for this selector.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -215,7 +215,7 @@ Do not infer and rebuild commands from project scripts. Bare React Native hosts
 Metro from the project's dependencies. Expo runs its fixed start command. iOS
 and Android use fixed `xcodebuild` and Gradle arguments.
 
-The supported build selectors are `ios --configuration <name>` and
+The supported build selectors are `ios --scheme <name>`, `ios --configuration <name>`, and
 `android --variant <name>`. `ios --device-type <name>`, `ios --runtime
 <version>`, and `android --system-image <id>` select the model and version of
 the owned simulator or emulator for one invocation, overriding
--- a/packages/core/index.ts
+++ b/packages/core/index.ts
@@ -180,6 +180,7 @@ export interface BuildRunOptions {
   variant?: string;
   abi?: string;
   configuration?: string;
+  scheme?: string;
   buildConfiguration?: string;
   isSimulator?: boolean;
   device?: string | boolean | null;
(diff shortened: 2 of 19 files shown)


janicduplessis:
Fresh review is clear at 1069e3bd80, including doctor recognition of scheme-specific warm products. All defined local checks, 67 end-to-end tests, and runtime-floor checks passed; required Node 22/24 and published Node 20 CI are green at this exact head. Merging under Janic’s overnight authorization.

PR #615, opened 2026-09-10
feat: report linked libraries whose Git metadata enters the fingerprint

### LINKED ISSUE #595: Linked native-library Git metadata prevents worktree artifact cache reuse
## Problem

Linked local native libraries can prevent artifact-cache reuse across Git worktrees even when native sources match. This occurred in the react-native-safe-area-context example, whose dependency is `link:..`.

## Evidence

The 1.0.0 candidate based on 5aa719fcfd built and launched the bare RN 0.85.3 iOS example successfully, but the warmed worktree missed the artifact cache (455/471 compiler-cache hits). Expo fingerprint debug output showed 231 differing files, all under `node_modules/react-native-safe-area-context/.git`: a directory in the main checkout versus a pointer file in the worktree.

A read-only fingerprint probe ignoring only that exact `.git` path and descendants produced equal fingerprints and identical 293-file sets. An intentional source edit still changed the fingerprint; restoring it restored equality. No native files were ignored.

## Cause

The installed Expo fingerprint hashes the symlinked library's Git metadata as part of its directory source. Creating a worktree changes the main checkout's Git metadata too.

## Fix idea

Document or diagnose a project-specific `.fingerprintignore` for linked library Git metadata when that project's native build does not read Git state. Avoid globally ignoring inputs that other builds may intentionally consume. This is a performance/setup finding, not stale artifact reuse.

janicduplessis:
## Description

`stim doctor` now names any `.git` inside a directory the native fingerprint hashes, such as a symlinked `link:` or `file:` library under `node_modules`, and prints the exact `.fingerprintignore` lines for it.

A linked dependency is a symlink into another checkout, so that checkout's `.git` lands in a directory `@expo/fingerprint` hashes whole. Git rewrites that metadata on every commit, checkout, or worktree of the linked checkout, so workspaces rarely compute the same fingerprint even when the library's native sources match, and every worktree compiles instead of reusing the artifact. Issue #595 measured this on the react-native-safe-area-context example: 231 differing files, all under the linked package's `.git`, and equal fingerprints once that one path was ignored.

## Solution

The check first looks for a symlinked package whose target holds a `.git` entry in every `node_modules` from the app directory up to the repository root, so hoisted workspace links count; that is a handful of `lstat` calls. only then does it run the project fingerprint once in debug mode and walk its `dir` sources, reporting a directory only when the fingerprint's own debug output shows it hashed a `.git` child. The printed path is therefore the one the fingerprint uses (the `node_modules` alias or a real path, whichever autolinking reported), packages that never enter the fingerprint are not mentioned, and a `.git` the project already excludes through `.fingerprintignore` or `fingerprint.config.js` is not reported, whatever pattern it used. It inspects the main checkout like the sibling doctor checks and honours `--platform`. `fingerprintProject` also passes `silent: true` now, because with `DEBUG` set `@expo/fingerprint` prints profiling lines to stdout, which would break `doctor --json`.

Two lines are printed per directory, `<dir>/.git` and `<dir>/.git/**/*`. The installed `@expo/fingerprint` matches ignore patterns per file and only skips a directory for patterns ending in `/**/*`, `/**` or `/`, so the first form covers the pointer file a worktree has and the second the directory a main checkout has.

**Doctor does not write the file.** The remedy holds only when the native build does not read Git state, which is the project's judgment, so the note says to ignore those entries alone, never the package. The guide's fingerprint section, the agent guide, and the commands page describe the finding.

## Test plan

- `doctor.test.ts` drives `detectLinkedLibraryGitMetadata` with a stub fingerprint whose debug output lists a scoped package that hashed `.git`, one that did not, and `android`. It asserts only the first is named with both ignore lines, that `--platform`, `debug` and `silent` reach the fingerprint call, the plural form with a real-path source, silence once no directory hashes `.git`, silence on a failing fingerprint, that no fingerprint runs when no `node_modules` holds a symlinked package with `.git`, and a hoisted link found from an app directory inside a repository.
- Ran the built CLI at `11c0ec9` against the real safe-area-context example with its linked native library. `doctor --platform android --json` reported only `linked-library-git-metadata` before the narrow ignore; after applying its exact printed `.git` and `.git/**/*` entries, findings were empty. Both outputs parsed as one JSON payload with `DEBUG=expo:fingerprint`. The disposable fixture was restored to its clean baseline afterward; no native build or original repository changed.

Fixes #595

### DIFF
--- a/packages/stim-cli/src/__tests__/doctor.test.ts
+++ b/packages/stim-cli/src/__tests__/doctor.test.ts
@@ -25,6 +25,7 @@ import {
   checkFingerprintParity,
   checkMetroCache,
   detectFingerprintParity,
+  detectLinkedLibraryGitMetadata,
   checkRemoteDevice,
   checkSimSlim,
   checkMainCheckout,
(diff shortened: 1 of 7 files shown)


janicduplessis:
CI note: GitHub stopped creating pull_request runs for this branch after the 03:39Z push (a reopen did not re-trigger them either). A workflow_dispatch run of ci.yml on the exact head commit 11c0ec9c9 passed: https://github.com/appandflow/stim/actions/runs/34435144431. A fresh push should re-attach checks to the PR.

janicduplessis:
Fresh independent code review at 11c0ec9 is clear: actual Expo ignore-matcher checks cover .git pointer files/directories and preserve native sources. One evidence correction before merge: the Trailhead run has no linked library, so the new gate skips fingerprinting; it cannot prove the fingerprint call runs inside doctor as the body says. Please correct that sentence and preferably run the built doctor on an affected linked-library fixture, showing the finding before the narrow ignore and no finding afterward. Do not overlap native builds on this Mac; doctor-only metadata is fine. Root app-config607 full checks now complete; PR forthcoming.

janicduplessis:
Codex completed the missing real affected-project check at11c0ec9: built doctor on the safe-area-context example reported the linked .git note before its exact ignore, then findings[] afterward. DEBUG=expo:fingerprint stayed off stdout, both outputs parsed as one JSON object. Disposable fixture restored clean. I replaced only the inaccurate Trailhead validation bullet with these observed results; source branch unchanged. Independent code review is clear. Proceeding with merge after exact-head required CI verification.

janicduplessis:
PR615 is currently merge-conflicted and has no exact-head checks, so it cannot merge yet. I am taking the integration follow-up in a separate local checkout: merge current main without rewriting your commits, resolve overlapping guide/config additions, rerun checks and fresh review, then push only as a normal fast-forward. No force-push; a concurrent remote update will be preserved and reconciled. The actual linked-library diagnostic is now verified and the body corrected.

janicduplessis:
CI clarification: the PR rollup was empty, but the exact 11c0ec9 head does have a successful manually dispatched CI run (34435144431). The merge conflict still requires an integration commit and new exact-head checks; no prior CI failure is implied.

janicduplessis:
Final independent integration review is CLEAR at b4718a68846e550e6fd586689cf26656cbcb2f9b. Current main config/scheme/reset changes are retained; the sole import conflict was resolved correctly. Reviewer checked the real before/after doctor evidence and single-JSON output. All local checks and required Node22/24/runtime20 CI passed. Merging this exact head.

PR #616, opened 2026-09-10
feat: reset Metro caches for one app

### LINKED ISSUE #606: Add an app-scoped Metro cache reset flow
## Problem

Stim has no explicit per-app Metro cache-reset recovery flow. Deleting a shared
cache globally is not equivalent to restarting one app with fresh Metro cache
state, especially with several live apps in a monorepo. The user approved adding
this capability after reviewing the Expo/RN CLI option audit.

## Evidence

Current main cc0cb528c8ecdf54619c393ee0fc084f05c2f374 exposes `start --json --wait
--remote`, without a cache reset selector. Expo/RN expose clear/reset options.
The ongoing real monorepo smoke has two independent Metro supervisors on ports
8082 and8083; a recovery action must preserve the other app and owned devices.

## Cause

Normal start is idempotent and reuses a healthy owned supervisor. There is no
request/state contract for a cache-reset restart, including Stim's injected shared
transform cache and the app's own Metro configuration.

## Fix idea

Add `stim start --reset-cache` with explicit owned-Metro restart semantics. Keep
the selected app's port and device, preserve other apps' running supervisors and
cache state, and refuse to take over an externally started Metro. Cover both Expo
and bare RN, new starts and existing starts, cache invalidation and restart
failure recovery. Document precisely what resets; do not promise isolation by
blindly clearing stores whose configured paths may be shared.

janicduplessis:
## Description

Recovering one app from stale Metro state should not clear caches used by other apps or shut down its device. Plain `start` currently reuses the server, so it cannot do that recovery. Fixes #606.

## Solution

`stim start --reset-cache` restarts only the app's verified owned Metro and preserves its port and devices. A saved cache generation bypasses old transform entries and selects a fresh file-map directory without deleting shared stores; ordinary starts reuse that generation. Starts and resets serialize per app, and unverified or externally started processes are left alone.

Expo requires SDK 54+ and the config adapter. Custom file-map cache managers must honor `fileMapCacheDirectory`; native caches are unchanged. Old entries remain on disk rather than being reclaimed by this command.

## Test plan

Ran real bare RN and Expo Metro servers together using a shared transform store. Reset each app in turn: its PID changed, its port stayed fixed, transforms reran, and the other app retained its PID and cached transforms. A subsequent ordinary restart reused the new cache; Metro wrote the new file-map cache. Smoke coverage was limited to Metro.

### DIFF
--- a/packages/stim-cli/README.md
+++ b/packages/stim-cli/README.md
@@ -45,6 +45,12 @@ errors in the captured logs. Exit code 0 alone means the query succeeded, even
 when it prints errors; an empty result does not prove launch or log capture
 succeeded.
 
+Use `stim start --reset-cache` to recover from stale Metro transforms or file-map
+state. It restarts only this app's verified owned Metro, keeping its port and
+devices. A fresh persistent cache namespace bypasses old entries without deleting
+shared stores or changing other apps or native build caches. Expo requires SDK
+54+ and Stim's config adapter. See `stim guide lifecycle` for scope and limitations.
+
 Use `stim doctor --platform ios` or `stim doctor --platform android` when only
 one native platform is in scope; shared project checks still run. Doctor also
 prints the running CLI version and the `stim` installation resolved from PATH,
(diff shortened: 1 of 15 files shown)


janicduplessis:
Fresh independent code review is CLEAR at1d644d57510776a39ca0fff94df59b1e7cf8f789, with129 additional focused tests. Full local checks and68e2e passed; real bare/Expo cache isolation and restart reuse proof passed. Required CI22/24/runtime20 are green at this exact head. Merging under Janic’s overnight authorization.

PR #635, opened 2026-09-10
fix: audit non-login benchmark shell commands

### LINKED ISSUE #632: Recognize non-login shell wrappers in benchmark audits
## Problem

A successful benchmark proof can be rejected as missing commands or missing agent-device session isolation when the runner uses a non-login shell.

## Evidence

The current parser in scripts/agent-benchmark/run-guards.mjs only unwraps /bin/zsh -lc (and bash/sh equivalents). The supported shell execution option login:false emits /bin/zsh -c instead. For example, `/bin/zsh -c 'env AGENT_DEVICE_STATE_DIR=/tmp/state AGENT_DEVICE_SESSION=run agent-device close'` remains wrapped even though the command and successful exit are captured. The viewer has a similar login-only normalization.

## Cause

The top-level shell wrapper regex requires -lc, so proof matching and command segmentation cannot see valid commands inside -c wrappers.

## Fix idea

Support the narrowly recognized -c and -lc shell forms without executing transcript text. Preserve exact standalone-command checks, quoting semantics, exit-code requirements, and rejection of extra shell suffixes. Add regression coverage and re-audit retained evidence rather than rerunning agents.

janicduplessis:
## Description

Benchmark audits reject successful proof commands when Codex uses a non-login shell (`/bin/zsh -c`), because only `-lc` wrappers are decoded. This also hides the explicit agent-device session from the isolation audit.

## Solution

Recognize `-c` alongside `-lc` for the existing supported shells. Keep quoting, standalone-command boundaries, and successful-exit requirements unchanged. This changes benchmark parsing only, not Stim or the measured agent workflow.

## Test plan

Regression cases cover non-login zsh/bash/sh proof commands, an unscoped chained command, and an extra shell suffix. The export fixture exercises complete proof validation through non-login wrappers.

Fixes #632.

### DIFF
--- a/scripts/agent-benchmark/run-guards.mjs
+++ b/scripts/agent-benchmark/run-guards.mjs
@@ -62,7 +62,7 @@ export function benchmarkTarget(config, selection) {
 
 export function topLevelShellCommand(command) {
   const trimmed = String(command ?? '').trim();
-  const match = trimmed.match(/^\/bin\/(?:zsh|bash|sh) -lc\s+([\s\S]+)$/);
+  const match = trimmed.match(/^\/bin\/(?:zsh|bash|sh) -l?c\s+([\s\S]+)$/);
   if (!match) return trimmed;
   const input = match[1];
   let source = '';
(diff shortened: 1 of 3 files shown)


janicduplessis:
Fresh independent review found no actionable findings and independently passed all 84 affected tests. Quoting, standalone command, session isolation, and successful-exit boundaries remain enforced. All defined local checks passed; exact-head required CI is green.

PR #637, opened 2026-09-10
fix: time actionable initial launch diagnostics

### LINKED ISSUE #636: Count actionable launch output in crash diagnosis timing
## Problem
The launch-crash report measures first actionable diagnosis at a later logs command even when the initial launch output already contains the injected error and symbolicated application source location. This overstates time to diagnosis.

## Evidence
A successful Android Stim run prints the crash token, `RootLayout (app/_layout.tsx:27:18)`, and the source code frame in the initial `stim android` response. Its diagnosis record nevertheless points to the following `stim logs --errors`. The run has valid repair, Settings screenshot, recording, and isolation evidence.

## Cause
At main 31b11414c2ceebf9edd67de5e44ce094570e4f39, `launchCrashDiagnosis` in scripts/launch-crash-benchmark.mjs requires `index > initialLaunchIndex` when choosing error capture, then starts actionable diagnosis search from that capture.

## Fix idea
Allow proven, actionable runtime error context in the launch response to establish diagnosis time, consistently across arms and platforms. Preserve the separate required log/proof/isolation contracts and reject echoed source or command text as runtime evidence. Recalculate from retained raw evidence without rerunning agents or altering measured commands.

janicduplessis:
## Description

Crash diagnosis timing ignores the initial launch response even when it already shows the runtime error and app stack, overstating time to diagnosis.

## Solution

Use actionable launch output when it arrives before the later diagnostic output, for both arms and platforms. Keep the separate successful log capture and source-inspection guards required for run validity; this changes derived timing, not the measured workflow.

## Test plan

Cases cover iOS/Android and Stim/control, missing or failed log capture, incomplete launch context, echoed source text, and overlapping commands whose logs finish first. Existing saved commands can be re-audited without rerunning agents.

Fixes #636.

### DIFF
--- a/docs/agent-benchmark.md
+++ b/docs/agent-benchmark.md
@@ -358,6 +358,9 @@ commands. The injected error text is unique per run so the
 collector can prove that the reported stack and repair refer to this failure.
 The unique token and source location must appear in captured runtime errors;
 the earlier successful launch command does not have to print the token inline.
+When that launch response already contains the runtime error and root-layout
+context, its completion establishes diagnosis time rather than a later log
+query. The separate successful log capture remains required for validation.
 Before capture, normal guide, doctor, worktree warming, and narrowly scoped
 installed-dependency resolution are setup, not application-source inspection.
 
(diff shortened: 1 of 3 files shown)


janicduplessis:
Fresh independent review found no actionable findings and independently passed all 38 launch-crash tests. All defined local checks and 83 combined focused tests passed; exact-head required CI is green. Saved-run read-only re-audit selects the initial launch response while preserving its separate successful logs evidence.

PR #642, opened 2026-09-10
fix: retain benchmark compatibility after cleanup

### LINKED ISSUE #638: Preserve benchmark compatibility evidence when recollecting after cleanup
## Problem
Recollecting a completed valid launch-crash run after successful cleanup marks it invalid solely because its worktree was removed. This prevents safe analysis-only timing corrections.

## Evidence
A run passed native compatibility, isolation, source-repair, screenshot and recording checks before cleanup. After normal cleanup, collect reports native-compatibility-changed-or-unverified with reason run worktree missing for compatibility validation. All events, transcript, proof, screenshot and recording hashes match the original valid record.

## Cause
The collector requires a live worktree for native compatibility verification, while durableRunRecord preserves source repair after verified cleanup but not that previously validated compatibility evidence.

## Fix idea
Retain the original validated compatibility evidence only after proven cleanup, matching run identity and unchanged captured evidence and compatibility manifest. Continue rejecting changed or unverified evidence. Add a regression test covering post-cleanup recollection and tampering. Do not waive native compatibility checks for new runs.

janicduplessis:
## Description

Recollecting a completed benchmark after cleanup rejects its native compatibility check because the worktree is gone. This blocked timing corrections even though the original run passed and its captured evidence was unchanged. Fixes #638.

## Solution

Retain the prior passing compatibility check only after recorded cleanup, matching run identity, unchanged captured evidence hashes, and a matching on-disk compatibility manifest. Missing worktrees remain invalid without that prior evidence. Keep the newly calculated timings and other audit failures rather than restoring the old verdict.

## Test plan

Replayed both retained Sol Android crash records against their post-cleanup recollections: each retains verified compatibility and the corrected diagnosis time, without rerunning the agent or changing recorded evidence. Regression cases cover changed identity, missing or changed evidence, changed manifests, previously unverified runs, and unrelated new audit failures.

### DIFF
--- a/docs/agent-benchmark.md
+++ b/docs/agent-benchmark.md
@@ -121,6 +121,11 @@ native compatibility adapter compensates. The policy, its digest, and the
 probe results are retained with the private run metadata; a refused run keeps
 its probe results in its run directory.
 
+Recollecting after recorded worktree cleanup can retain a previously passing
+compatibility check when run identity, captured evidence hashes, and the pinned
+manifest are unchanged. Other audit checks and derived timings are recalculated;
+missing or changed evidence does not inherit the earlier verdict.
+
 This boundary prevents accidental benchmark-data access, not adversarial host
 access: native system services and pre-existing processes are not sandboxed by
 the runner policy. Strong isolation from a malicious agent requires a separate
--- a/scripts/agent-benchmark/native-compat.mjs
+++ b/scripts/agent-benchmark/native-compat.mjs
@@ -211,7 +211,15 @@ export function collectedNativeCompatibility(meta, worktree) {
   try {
     if (!meta.preflight.nativeCompatibilityProbe?.processIdentity)
       throw new Error('sandbox compatibility probe missing');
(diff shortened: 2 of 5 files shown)


janicduplessis:
Fresh independent correctness review is clear for b4a313c912daed63199e930583fe48849daa7eb9 after tracing collection, worktree evidence, source proof, manifest verification and cleanup gating. Description skimmer also clear. All local format/lint/build/typecheck/3835 tests/78 e2e/runtime-floor/knip checks passed before the timed benchmark resumed. Review was static to avoid competing with that benchmark.

janicduplessis:
Node 22 CI failed in the unchanged start.test.ts remote-supervisor test (line2162): expected STIM_REMOTE_START_REQUIRED, got STIM_METRO_TIMEOUT. The test starts a real listener asynchronously and gives runAction a one-second wait; Node24 and local suites passed. Investigating this timing-sensitive test separately and rerunning only the failed job at the same reviewed SHA; no assertion or cache-audit guard has been relaxed.

Issue #644, opened 2026-09-10
Report compiler cache statistics before a later launch failure


janicduplessis:
## Problem
A native Android build succeeds but its compiler-cache statistics disappear from CLI output if a later install or launch step fails. A retry then reports artifact-hit/not-run, obscuring the cache performance of the original build.

## Evidence
A pooled Android run compiled in28.1s, installed its APK, then failed adb reverse because the device was offline. The retry launched successfully from the artifact cache. The per-build ccache-stats.log contains386 hits and0 misses, but neither command prints those counters. The benchmark flagged missing evidence as intended.

## Cause
commands/android.ts receives built.ccache after build completion, but the human compilation-cache summary is emitted in android/result.ts at the successful end of the launch workflow. A pre-summary launch failure bypasses it.

## Fix idea
Emit or persist compiler-cache results at build completion, including failed-build paths where counters are available, and keep JSON failure facts useful. Avoid duplicate success-summary noise. Verify that a later launch failure cannot erase an already completed build measurement. Check analogous iOS behavior too.

janicduplessis:
Claimed: @janic/issue-644-cache-stats-before-launch

PR #646, opened 2026-09-10
fix: preserve benchmark compiler cache evidence

### LINKED ISSUE #645: Capture per-build ccache evidence independently of launch summaries
## Problem
An otherwise valid recovered Android native benchmark is rejected because a later launch failure prevents the CLI from printing the successful build compiler-cache counters. The retained per-build stats log proves the counters, but benchmarkCcache only reads command output.

## Evidence
Native build28.1s, then adb reverse fails with device offline. Retry succeeds from artifact cache. A single native compilation occurred; its workspace ccache-stats.log records386 hits,0 misses. The existing audit correctly refuses absent command-output counters, so simply waiving ccache-evidence-missing is not appropriate.

## Cause
The benchmark does not preserve and bind the workspace per-build CCACHE_STATSLOG to a specific completed build before cleanup. Stim resets that log before each compile, so only the last build survives; it must not be reused for earlier attempts.

## Fix idea
Capture immutable hash-bound per-build stats evidence and audit it independently from launch output. Require exact run/workspace/command provenance, distinguish a single measured build from ambiguous multiple attempts, reject stale or modified evidence, preserve low-hit alerts, and retain the raw log before cleanup. Do not alter recorded command output or label unmeasured builds healthy.

janicduplessis:
## Description

A native benchmark compiled successfully, then failed during launch before Stim printed its compiler-cache counters. The agent recovered, but the audit rejected the run despite its per-build log proving 386 hits and no misses. Fixes #645; #644 tracks improving the CLI output separately.

## Solution

Capture the run's ccache log only for one unambiguous completed build whose command lacks counters. Require the isolated workspace, a log modification time inside that command, and hashes binding the capture to its metadata, events and command. Recollection and website export verify the capture; missing evidence and low-hit alerts still apply. Multiple builds remain unsupported because Stim resets the log on each build.

## Test plan

Replayed the retained Luna native commands and original timestamped stats log in a separate temporary capture: 386/386 hits, above the 95% target, without modifying the recorded commands. Regression cases reject stale, ambiguous, modified, missing, or misattributed evidence and retain low-hit alerts. Website export rejects a record claiming log-based counters without its capture. No published CLI or benchmark prompt changes.

### DIFF
--- a/docs/agent-benchmark.md
+++ b/docs/agent-benchmark.md
@@ -126,6 +126,12 @@ compatibility check when run identity, captured evidence hashes, and the pinned
 manifest are unchanged. Other audit checks and derived timings are recalculated;
 missing or changed evidence does not inherit the earlier verdict.
 
+If a later launch failure omits compiler-cache counters, collection can preserve
+the workspace's per-build ccache log for one unambiguous completed Android build.
+Its modification time must fall within that command, and saved hashes bind it to
+the run, metadata, events, and command. Recollection and website export verify
+those hashes; multiple builds, stale logs, and low hit rates still need investigation.
+
 This boundary prevents accidental benchmark-data access, not adversarial host
 access: native system services and pre-existing processes are not sandboxed by
 the runner policy. Strong isolation from a malicious agent requires a separate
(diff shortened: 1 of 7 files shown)


janicduplessis:
Fresh independent review is clear at b12c121f18540b3b8654e90313691be35c84798f: attribution, timestamp/hash checks, cleanup survival, cache guards, exporter reconstruction, tests and documentation reviewed. Reviewer independently ran 62 focused evidence/export tests successfully. Local full checks and exact-head Node22/24/runtime20 CI are green.

PR #649, opened 2026-09-10
fix: clear completed recording state before parking

### LINKED ISSUE #648: Preflight pooled Android recording state before timed benchmarks
## Problem
A timed Android Stim launch-crash run recovered the app but could not record proof because an earlier completed agent-device recording manifest survived on the pooled emulator. The old PID was reused, causing recording-start refusal before any new video could begin.

## Evidence
agent-device0.20.10: Android screenrecord completed evidence cannot be safely retired. Completed manifest names pid4004/startTime3766; current proc stat is an unrelated binder process/startTime5432. Remote old video is absent and original captured proof is retained. Expected Stim-owned parked AVD was adopted. No benchmark command or cache issue caused the recording failure.

## Cause
Pooled device storage preserves agent-device completed native recording manifests; the harness preflight does not test or retire this completed state. Upstream inspect reports ownership-lost for PID reuse and retirement accepts only missing.

## Fix idea
Audit and prepare recording state outside timed runs on the exact verified owned emulator, preserving evidence before cleanup. Retire only schema-valid completed records whose original recorder is proven gone and whose saved output exists; never kill unrelated processes or waive recording validation. Add regression coverage for reused PIDs, live/uncertain identities, incomplete manifests, and wrong-device/session evidence. Report upstream separately and verify actual record start/stop on the retained emulator before retrying the failed run.

janicduplessis:
## Description

A pooled Android emulator retained a completed agent-device recording manifest. After its PID was reused, the next benchmark could repair the app but could not start its required video. Fixes #648; upstream report: https://github.com/callstack/agent-device/issues/2476.

## Solution

Retire completed recording metadata during benchmark cleanup, before stopping and parking the emulator. Require the owned device, matching saved-video hash, absent remote chunks, and proof that each original recorder is gone. Archive the manifest and delete only that metadata; never signal the replacement process. Unknown or changing evidence stops cleanup. This changes neither the published CLI nor agent-device, and does not preboot the next run's emulator outside timing.

## Test plan

On the retained Android36 emulator, verified the old PID belonged to a different process with a different start time and the old video was absent. The helper archived and removed the completed manifest; a separate diagnostic session then successfully started and stopped a new recording on the same emulator. Regression cases refuse live/unreadable identities, incomplete/foreign manifests, altered video proof, remaining chunks, and ownership/manifest changes.

### DIFF
--- a/docs/agent-benchmark.md
+++ b/docs/agent-benchmark.md
@@ -355,6 +355,15 @@ control, remove its worktree and branch and shut down and delete only the newly
 created benchmark simulator or emulator. Do not touch unrelated physical-device
 leases or automation sessions.
 
+Before stopping a pooled Android emulator, cleanup also retires completed
+agent-device recording manifests. It requires the exact owned device, the
+matching run's hash-verified saved video, absent remote video chunks, and proof
+that every original recording process is gone (including a different process
+start time after PID reuse). It archives the manifest and removes only that
+metadata; it never signals the replacement process. Unknown, incomplete, live,
+or changed evidence stops cleanup. This runs after timing, not by prebooting a
+parked emulator before the next agent starts.
+
 ## JavaScript launch-crash extension
 
 A launch-crash diagnostic is a separate suite, not a fifth performance cell.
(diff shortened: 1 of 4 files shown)


janicduplessis:
Fresh independent correctness review is clear at d8a77efadd. Ownership/proof checks, PID reuse versus live or uncertain identity, and cleanup-before-parking integration reviewed; all 16 focused regression tests passed independently. Real emulator recovery and fresh recording start/stop passed. Full local checks and exact-head Node22/24/runtime20 CI are green.

PR #650, opened 2026-09-10
fix: identify real benchmark launch commands

### LINKED ISSUE #647: Do not count CLI help as benchmark launch evidence
## Problem
The launch-crash audit records an Expo help query as the initial launch. A later genuine build/install/launch succeeded, so this recovered run is valid in practice, but the initial-launch command metadata is wrong.

## Evidence
On current main 2acde6e534, a completed `npx expo run:android --help | sed -n ...` with exit0 is selected ahead of the actual `set -o pipefail; adb -s emulator-5554 reverse ... && ORG_GRADLE_PROJECT_reactNativeArchitectures=arm64-v8a npx expo run:android ... 2>&1 | tee ...` command. The latter completed successfully before device log diagnosis and source repair. No application source was inspected before diagnosis.

## Cause
scripts/launch-crash-benchmark.mjs launchCommand recognizes the command name without excluding help/version invocations. successfulLaunch additionally rejects the legitimate pipefail-protected launch pipeline containing a preceding scoped adb reverse via its delimiter restriction.

## Fix idea
Reject non-executing help/version queries as launch proof and recognize the observed safe compound launch using parsed shell segments and propagated status. Preserve strict failure/pipeline-status and source-before-diagnosis checks. Regression-test both cases, re-audit retained evidence rather than rerunning the agent, and update only analysis provenance before publication.

janicduplessis:
## Description

The launch-crash audit could label `expo run:android --help` as the initial launch while rejecting the real launch preceded by `adb reverse`. This affects benchmark evidence, not Stim runtime behavior.

## Solution

Exclude help/version invocations and accept device-scoped Android port forwarding chained to a launch in a successful `pipefail` pipeline. Failed or status-masking commands remain rejected.

## Test plan

Replayed a retained Android control run: the initial-launch event now points to its real Expo invocation instead of help; the diagnosis event and timestamp are unchanged. Regression cases cover help flags, forwarding, failed launches, masked status, and intervening source reads.

Fixes #647.

### DIFF
--- a/scripts/launch-crash-benchmark.mjs
+++ b/scripts/launch-crash-benchmark.mjs
@@ -61,7 +61,9 @@ function successfulLaunch(command, arm, platform) {
     if (statuses.length !== 1 || statuses[0][1] !== '0') return false;
     pipeline = pipeline.slice(0, report.index);
   }
-  const [launch, ...filters] = pipeline.replace(/\s+2>&1(?=\s|$)/g, '').split(/\s*\|\s*/);
+  let [launch, ...filters] = pipeline.replace(/\s+2>&1(?=\s|$)/g, '').split(/\s*\|\s*/);
+  if (arm === 'control' && platform === 'android')
+    launch = launch.replace(/^adb\s+-s\s+emulator-\d+\s+reverse\s+tcp:\d+\s+tcp:\d+\s*&&\s*/, '');
   const group = /^\{\s*([\s\S]*?);\s*\}$/.exec(launch);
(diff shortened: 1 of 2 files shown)


janicduplessis:
Fresh independent correctness review is clear at 678fb15477708e628559700b63331f6379d3df5e: guarded forwarding, failure propagation, and source-before-log checks reviewed; 38 focused tests passed. Description review: 5/5, no cuts. Full local checks passed, including 3,852 unit tests and 78 E2E tests. Retained-run replay selects the genuine initial launch and preserves the diagnosis event/time.

PR #657, opened 2026-09-10
feat: refresh the main checkout before warming, under a lock

### LINKED ISSUE #651: worktree warm --refresh: refresh the main checkout before copying, under a lock
## Why

A warm worktree is a copy of the main checkout, so whatever is stale there is stale in every worktree it seeds. `doctor` already reports how far behind the main checkout is and whether `node_modules` and `ios/Pods` match their lockfiles, but fixing that is left to the user: pull, install, maybe `pod install`, by hand, in the main checkout, before every `warm`.

With several agents on one machine that hand-rolled refresh is unsafe in a way nothing reports:

- Two `git pull`s collide on `index.lock`. The second fails; harmless but noisy.
- Two `pnpm install`s in one `node_modules` are not safe.
- Worst: one agent's `stim worktree warm` copies a `node_modules` that another agent's install is halfway through. The worktree is silently broken, and nothing that ran says so.

Stim is the right owner because `warm` is the reader of that state. The refresh belongs beside the copy, under a lock, so the copy can never observe a half-refreshed main checkout.

## `stim worktree warm --refresh`

The refresh is opt-in, and it is opt-in for a reason that outranks taste: today `warm` works fine against a dirty main checkout, so refusing on a dirty tree by default would break every existing user with uncommitted work in main. The flag also marks the surprising part: this writes to a checkout other than the one you are standing in.

With `--refresh`, before the copy:

1. **Refuse a main checkout it cannot move.** Uncommitted tracked changes, a rebase or merge in progress, or a detached HEAD each stop the command with the paths or the state named and the exact git line that clears it (`git stash push -u -m ...` or a commit for a dirty tree; `git checkout <branch>` for a detached one). Untracked files are not a reason to refuse. Error codes: `STIM_MAIN_DIRTY`, `STIM_MAIN_DETACHED`.
2. **Fetch and fast-forward whatever branch is checked out.** `git fetch --prune` the remote, then fast-forward the main checkout's branch to its `@{upstream}`. A branch with no upstream is left where it is and said so. A branch that has diverged (ahead and behind) stops with `STIM_MAIN_DIVERGED` rather than merging or resetting. Already up to date is a normal, quiet outcome. A fetch that fails (offline) is reported as a fact and the run continues with the local state; the network is not a reason to refuse.

   `--refresh` never switches branches. Fast-forwarding what is already checked out is safe on any branch, and switching a branch under someone else's checkout is not a thing a copy command should do.
3. **Install only what moved.** If the lockfile changed between the old and new HEAD, or `node_modules` does not match it (the `depsOutOfSync` check doctor runs), run the install command Stim already maps from the lockfile (`pnpm install`, `yarn install`, `bun install`, `npm ci`) in the main checkout. Otherwise skip and say so.
4. **Pods likewise.** If `ios/Podfile.lock` changed or `ios/Pods/Manifest.lock` does not match it (`podsOutOfSync`), run `pod install` the way `stim ios` would (bundler when a Gemfile is present). Skip when there is no `ios/`, and say so when skipped.

Then the copy runs exactly as it does today.

## Warn when the main checkout is not on the default branch

Fast-forwarding a feature branch is safe, but it is rarely what someone wants from a checkout whose whole job is to seed worktrees: the copy then carries that branch's dependencies into every worktree made from it. So `--refresh` warns, and continues.

Resolve the default branch in this order, stopping at the first that answers:

1. `worktree.defaultBranch` in the repository-root `.stim.json` (same scope as `worktree.exclude`, which is already repository-wide rather than per-app).
2. `git symbolic-ref --short refs/remotes/origin/HEAD`, minus the remote prefix.

If neither answers, emit no warning and say why once, with the line that fixes it (`git remote set-head origin -a`). An unresolvable default branch is not a refusal and not an error; it just means this particular warning cannot be made.

Nothing else in Stim reads `worktree.defaultBranch`; it exists so a repository whose `origin/HEAD` is missing or wrong can still get the warning right.

## The lock, which is not opt-in

One lock per main checkout, PID-held and stale-safe like the build locks: a holder that dies frees it.

- `--refresh` holds it **exclusively** for the refresh phase.
- Every `warm`, with or without the flag, holds it **shared** for the duration of its copy.

Plain `warm` must take the shared hold even though it mutates nothing: without it, one agent's copy can read a `node_modules` that another agent's `--refresh` is rewriting, which is the corruption this issue exists to stop. Two plain warms proceed together. Waiting prints a progress line every 30s in the same shape as build waits (`lock  waiting on stim worktree warm --refresh (pid 41233, 40s elapsed)`), and gives up after the same kind of bound the build wait uses, naming the holder.

**Key the lock on the REPOSITORY root, not the project root.** Stim commands run from an app directory, so the obvious `findProjectRoot` is the wrong key here: in a monorepo it would give `apps/a` and `apps/b` separate locks, and two agents warming different apps would both fast-forward the same git repository and both run `pnpm install` into the same root `node_modules`. Warm is already repository-wide (`guide settings`: "Worktree copying is repository-wide"), and everything the refresh touches -- the git checkout, the lockfile, `node_modules` -- is repository-wide too. Use the main checkout's canonical repository root, the same value `repoRoot()` resolves.

A plain `git worktree add` copies nothing and takes no lock.

## Monorepos

The refresh runs at repository scope, with one exception:

- **Deps** install where the lockfile is, which is the repository root, not the app directory.
- **Pods** are per-app, so `--refresh` runs `pod install` for the app the command was invoked from, and names it in the `pods` line. Another app's Pods in the same repository stay as they are until a `--refresh` runs from that app. Do not walk the repository installing pods for every app.

A repository with no `ios/` under the invoking app skips the pods step and says so.

## Output

Fact lines, one per line, in the same voice as `ios`/`android`, before the existing copy output:

    lock        acquired (waited 12s for stim worktree warm --refresh pid 41233)
    checkout    develop 3 commits behind origin/develop -> fast-forwarded to 9f2c1a3
    deps        pnpm-lock.yaml changed -> pnpm install (41s)
    pods        apps/tlon-mobile: ios/Podfile.lock unchanged -> skipped

On a branch that is not the default one, the `checkout` line carries the warning:

    checkout    janic/wip 2 commits behind origin/janic/wip -> fast-forwarded to 4b81e0c
                not the default branch (develop); worktrees seeded from this copy
                carry janic/wip's dependencies

Without the flag, only the `lock` line appears, and only when the copy actually waited.

After the copy, if this worktree's lockfile differs from the one whose `node_modules` was just copied (the worktree is based on a different commit than main), say so in one line with the install command to run here. Nothing in `warm` builds anything.

## Everything that mentions the gap should point here

- The doctor finding `The main checkout is N commits behind` names `stim worktree warm --refresh`; so do the `node_modules` / Pods out-of-sync findings.
- `guide lifecycle` shows the flag in the worktree step and says what it does to the main checkout, and `guide errors` documents the three error codes.
- The skill mentions it in one line where it describes the main checkout.

## Tests

- Pure: the decide-what-to-run logic over (old HEAD, new HEAD, lockfile state, pods state) without touching git.
- Real git: a temp repo with a bare remote, main behind by N commits, `--refresh` fast-forwards it; a diverged main refuses; a dirty main refuses and names the path; a detached main refuses; a main with no upstream is left alone; **a dirty main with a plain `warm` still copies, unchanged from today.**
- The lock: a `--refresh` in flight makes a plain `warm` wait; two plain warms do not block each other; a dead holder frees it; **two apps in one monorepo resolve the same lock key**, and a `--refresh` from `apps/a` blocks a `warm` from `apps/b`.
- The default-branch warning: fires on a feature branch, silent on the default branch, honours `worktree.defaultBranch` over `origin/HEAD`, and says why it cannot warn when neither resolves.
- Guide contract test for the new flag, the new setting, and the error codes.

## Out of scope

Building (`warm` never runs `ios`/`android`), any write to a linked worktree beyond the copy, and merging or resetting a diverged branch.

janicduplessis:
## Description

`stim worktree warm --refresh` brings the main checkout up to date before the existing copy:
fetch, fast-forward whatever branch is checked out there, install what the new commits moved.
Today that is hand work before every warm -- `doctor` reports the gap (commits behind, stale
`node_modules`, stale Pods) and leaves the fixing to the user.

**This changes plain `warm` too:** every warm now takes a *shared* hold on a per-repository lock
for its copy, and `--refresh` holds that lock exclusively. Two plain warms never block each other,
but a plain warm can now wait behind a running `--refresh`. That is the point: with several agents
on one machine, one agent's `warm` otherwise copies a `node_modules` that another agent's install
is halfway through, and nothing that ran says so.

Three further behaviours are here because each is a way this feature could silently hand back the
broken worktree it exists to prevent:

- **A present `node_modules` is not proof of a successful install.** A `--refresh` whose install
  failed leaves the directory there with tracked files clean, so on the retry HEAD no longer moves,
  directory existence counts as installed, and the run reported `pnpm-lock.yaml unchanged ->
  skipped`, copied partial dependencies and exited zero. The refresh now records the lockfile hash
  of a *completed* install under `~/.stim/warm-installs` -- written incomplete before the install,
  completed after it -- and reinstalls when that evidence is missing or stale. A repository with no
  record behaves as before, so nobody is forced into a reinstall on upgrade.
- Run from an app directory upstream deleted, `--refresh` stopped at `no lockfile in this
  repository` and copied the old dependencies. `dependencyState` now falls back to the repository
  root.
- The dirty check reads the index as well as the working tree. `git diff HEAD` is empty for a
  staged edit whose working file was restored to the HEAD contents -- `git status` calls that `MM`
  -- so `--refresh` used to fast-forward a dirty checkout instead of returning `STIM_MAIN_DIRTY`.

Closes #651.

## Solution

Pure decision logic (`checkoutPlan`, `depsPlan`, `podsPlan`, `mainCheckoutRefusal`,
`defaultBranchNote` in `src/worktree-refresh.ts`) is unit-tested without git; the I/O around it
reuses what exists -- `dependencyState` plus the lockfile-to-install-command table,
`hasInstalledDependencies`, `installedNpmTreeIsValid` and `locallyKnownUpstream` from `doctor.ts`
(now exported), and `readPodState` / `podsAreStale` / `runCaptured` / `runPodInstall` from
`engine/deps.ts`.

The flag is opt-in for compatibility, not taste: plain `warm` works fine against a dirty main
checkout today, so refusing one by default would break every user with uncommitted work in main.
With the flag:

- Refuses a main checkout it cannot move: `STIM_MAIN_DIRTY` (tracked changes, a rebase, `git am` or
  merge in progress, a status git could not report at all, or a fast-forward git itself refuses),
  `STIM_MAIN_DETACHED`, `STIM_MAIN_DIVERGED` -- each naming the git line that clears it.
- Untracked files are not a reason to refuse; a branch with no upstream is left where it is; a
  failed fetch is reported as a fact and the run continues on local state.
- Never switches branches. It only WARNS when the checked-out branch is not the default one,
  resolved from `worktree.defaultBranch` in the repository-root `.stim.json`, else `origin/HEAD`.
  Neither resolving means no warning and one line saying why, not an error.
- Monorepo scope: dependencies install where the lockfile is (the repository root); `pod install`
  runs only for the app the command was invoked from and is named in the `pods` line. Nothing
  walks the repository installing pods for every app.

The lock (`src/engine/warm-lock.ts`) is a PID-held reader/writer lock under
`~/.stim/warm-locks`: a holder that dies is reaped, a wait heartbeats every 30s and gives up with
`STIM_LOCK_TIMEOUT` naming the holder. It is keyed on the main checkout's **repository root**, not
`findProjectRoot`: in a monorepo the project root would give `apps/a` and `apps/b` separate locks
while both fast-forward the same git repository and both install into the same root
`node_modules`. The `git fetch` is bounded and runs with `GIT_TERMINAL_PROMPT=0`, because a
credential prompt behind the exclusive lock would block every other warm in the repository.

Two exclusive holders at once is the failure this lock exists to prevent, so its primitives are
chosen so that no interleaving produces one: the writer claim is built in a private directory and
renamed into place (`rename` onto a directory that already holds a record fails, so a claim is
never visible half-built and no mtime grace period is needed anywhere); a claim's token is the name
of its own record file and is read together with its holder, so reaping is an `unlink` of exactly
the claim that was observed and can never take the live one that replaced it; the claim carries the
installer child's pid beside Stim's, because the install runs in a process that can outlive Stim
and keep writing `node_modules`; and a reap the filesystem refuses throws instead of leaving the
dead claim in place for every later refresh to wait out the ceiling.

Plain `warm` must keep working wherever it worked before the lock existed, so a copy that cannot
take the lock at all -- a read-only `STIM_HOME`, say -- prints one dim `lock unavailable` line and
copies unsynchronised. `--refresh` refuses instead, because it genuinely needs the lock. A
`STIM_LOCK_TIMEOUT` is not that case and still refuses: a refresh really is holding the repository.
A copy also reads its exclusion settings after it holds the shared lock, so a refresh that lands
while it waits changes what it carries.

## Test plan

- `pnpm format:check`, `pnpm lint`, `pnpm typecheck`, `pnpm knip` (clean), `pnpm test` --
  123 files, 3902 passed, plus `pnpm run test:e2e` (78 passed).
- Backward compatibility for plain `warm`, measured rather than argued: the same seven scenarios
  (clean, dirty, detached, diverged, untracked plus a nested ignored directory, a 0500 `STIM_HOME`,
  and a `worktree.exclude`) run against a CLI built from `origin/main` and from this branch produce
  byte-identical stderr and exit codes, except for the one added `lock unavailable` line in the
  0500 case.
- `src/__tests__/worktree-refresh.test.ts` drives the real command against a real git fixture
  (bare remote + main clone + linked worktree): a behind main is fast-forwarded then copied; a
  diverged, dirty, and detached main each refuse and leave HEAD alone; **a dirty main with a plain
  `warm` still copies exactly as before**; a branch with no upstream is left where it is; an
  unreachable remote reports the fetch and continues; the default-branch warning fires on a
  feature branch, stays silent on the default one, honours `worktree.defaultBranch` over
  `origin/HEAD`, and says why when neither resolves. A monorepo case asserts `pnpm install` runs
  at the repository root and `pod install` only for the invoking app, and that a `--refresh` in
  flight blocks a plain warm started from a *different* app of the same repository. It also drives
  the three behaviours above end to end: a failed install then a retry that reinstalls (`the last
  install of pnpm-lock.yaml did not finish -> pnpm install`) then a third run that skips; a
  `--refresh` from an app upstream deleted installing at the repository root; a staged edit whose
  working file is back at HEAD refusing; the installer pid on the claim while the install runs and
  gone after it; a plain warm copying with a 0500 `STIM_HOME` while `--refresh` refuses; and a copy
  that waited picking up a `.stim.json` exclusion written during the wait.
- `src/__tests__/warm-lock.test.ts` covers the lock directly: two copies hold it together, two
  refreshes never do, a copy waits for a refresh and names it, a refresh drains live copies first,
  dead holders on both sides are reclaimed without waiting, and the ceiling refuses with
  `STIM_LOCK_TIMEOUT`. One case per interleaving above: a claim swapped for a live one under a
  reader's own liveness check survives that reader's reap; a writer directory holding no record is
  taken without waiting and the claim that lands in it is complete; a dead Stim pid with a live
  installer pid still makes a copy wait; and a writer directory a reaper cannot unlink inside (mode
  0500) rejects with `EACCES` rather than waiting.
- Real-tool check (invariant 9): the git calls run against real git in those fixtures, and
  `--refresh` ran a real `npm ci` end to end in a scratch repo
  (`deps        no installed dependencies -> npm ci (164ms)`), recording
  `{"lock":"package-lock.json","hash":"07107fba...","completed":true}`. A manual monorepo run:

```text
  lock        acquired
  checkout    janic/wip 1 commit behind origin/janic/wip -> fast-forwarded to e4e9832
              not the default branch (main); worktrees seeded from this copy
              carry janic/wip's dependencies
  deps        pnpm-lock.yaml unchanged -> skipped
  pods        apps/mobile: ios/Podfile.lock unchanged -> skipped
  carry       copied .env2 from /tmp/demo/main
  carry       complete: 1 ignored entries copied, 2 kept, 0 failed
```

Rollback is a revert; nothing persists outside `~/.stim/warm-locks` and `~/.stim/warm-installs`,
both safe to delete (deleting the install records only costs one extra install per repository).


janicduplessis:
Second independent review (Codex, gpt-6-astra) of the **rewritten** lock: eight more defects, four P1, each reproduced — several with real subprocesses, including killing a live `npm ci` mid-postinstall. I confirmed #4 by reading the code.

Holding this PR rather than starting a third fix round, because the pattern is a design problem, not an effort problem.

**Four of the eight are the same shape**, and it is inherent to the primitive:

1. B reads A's record showing `installerPid: null`; A then records installer C and dies; B acts on its stale snapshot and unlinks a claim that was protecting a live C.
3. The installer child starts *before* its PID is published, so a kill in that window leaves `installerPid: null` on a claim whose installer is alive.
8. A recycled PID keeps a dead claim unreapable until an unrelated process exits.

All three are the gap between "this PID exists" and "the record says what it is doing". You cannot close that gap by writing the record more carefully, because the process can die between any two writes. Finding 2 is a different fundamental hole in the same area: tracking the direct installer does not cover its descendants, and a package manager's postinstall child keeps writing `node_modules` after the parent is killed.

**The correct primitive is kernel-enforced liveness** — an advisory lock held by an open file descriptor, released by the OS when the holder dies, with no PID polling, no reaping and no TOCTOU; plus a process group rather than a single child PID so descendants are covered. That is a different design from this one, not a patch to it.

**Why holding beats shipping.** This lock exists to stop a copy reading a half-written `node_modules`. A lock with these holes still allows exactly that, while making callers believe it cannot happen. No lock is honest; a broken lock is worse than none.

Also worth recording, because it is mine rather than the implementation's: finding 4 is my instruction followed too literally. I asked for a copy that cannot take the lock to degrade to an unsynchronised copy, so plain `warm` would keep working on an unwritable `STIM_HOME`. `commands/worktree.ts:185` applies that fallback with no `opts.refresh` check, so a `--refresh` whose reader-record write fails also proceeds unlocked. The degradation belongs only on the unflagged path.

The rest of the change — the fast-forward, the refusals, the default-branch warning, the install-completion evidence, the monorepo scoping — is sound and independently verified. It is the cross-process lock that is not ready.

Full findings are in the review; the design issue is filed separately.

janicduplessis:
Superseded by #690, which implements the same `worktree warm --refresh` on the shared ownership-claim primitive from #681 instead of a lock of its own.

The feature in this PR was sound; its hand-rolled lock was not. Across four review rounds it accumulated 25 reproduced defects, all of the same few shapes -- a record and a process read at two different times, PID existence mistaken for process identity, and one PID standing in for a process tree. #678 extracted the primitive those defects motivated, #681 migrated the three locks that had each rolled their own, and #690 makes warm its first new consumer.

Everything else here carried over: the fast-forward and its refusals, the default-branch warning, the install ledger, the monorepo scoping.

PR #659, opened 2026-09-10
fix: pin benchmark locale for native cache reuse

### LINKED ISSUE #656: Pin benchmark locale so runners share compiler cache keys
## Problem

An Android native Stim benchmark recorded 18 compiler-cache hits and 368 misses, despite using the same golden cache that earlier runners reused at 100%. Preparation and agent shells do not share a pinned locale.

## Evidence

With published Stim 1.0.0 and ccache 4.13.6, a read-only replay of the exact Ninja invocation for one generated autolinking translation unit hits the failed run's own cache with `LANG=en_US.UTF-8`, `LC_ALL=en_US.UTF-8`, and `LC_CTYPE=C.UTF-8`, but misses the golden cache. Changing only `LANG` and `LC_ALL` to `C.UTF-8` produces a direct hit against that golden cache. The object and dependency-file state were restored after each probe. PCH was disabled, and the benchmark changed only the Android app label.

## Cause

ccache's debug input includes LANG, LC_ALL and LC_CTYPE in the common cache key. Benchmark preparation and runners inherit locale variables rather than enforcing the same environment, producing incompatible keys without a source change. The locale mismatch is proven for the replayed unit; broader verification is still required.

## Fix idea

Pin the same locale for golden preparation, runner processes and agent shell commands in both arms. Record and check it as benchmark provenance. Preserve the original failed attempt, verify native cache reuse under the pinned locale, then retry. Do not change Stim's runtime locale or weaken cache validation to fix a harness mismatch.

janicduplessis:
## Description

An Android native benchmark fell from full compiler-cache reuse to 18 hits / 368 misses because its inherited locale differed from preparation. ccache includes locale variables in its keys, even when the source and compiler are unchanged.

## Solution

Pin the same locale for preparation and both agent arms, including login and non-login shells. Record it in preparation/run metadata and refuse golden reuse without matching provenance. This changes benchmark infrastructure only, not Stim's runtime or cache validation thresholds.

## Test plan

On the benchmark Mac, replayed exact Ninja invocations with read-only cache access. The generated autolinking unit missed the golden cache with `en_US.UTF-8` and directly hit with `C.UTF-8`; a second native unit (`OnLoad.cpp`) also directly hit with the pinned locale. Original generated objects were restored. Real-shell regression tests verify locale and PATH restoration despite conflicting inherited settings.

Fixes #656.

### DIFF
--- a/docs/agent-benchmark.md
+++ b/docs/agent-benchmark.md
@@ -52,6 +52,13 @@ and no unexpected listener on Metro ports 8081 through 8090. Wait for a
 one-minute load average at or below 3.0 for two consecutive 15-second samples.
 Runs are sequential.
 
+Preparation and both agent arms pin `LANG`, `LC_ALL`, and `LC_CTYPE` to
+`C.UTF-8` and unset `LC_MESSAGES`. These variables participate in ccache keys:
+inheriting a different runner locale can turn a warm native build into misses.
+The isolated shell restores the locale for login and non-login commands.
+Preflight checks it, and preparation/run metadata records it. An older golden
+without matching locale provenance needs requalification before dispatch.
+
 Each machine has targets for every platform, change kind, and arm. The
 screen-ready target is a visible performance signal, not a validity gate,
 because agent reasoning time is part of what the benchmark measures. A Stim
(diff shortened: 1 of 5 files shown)


janicduplessis:
Fresh correctness review found a retained iOS preparation path that could relabel an older cache. It now requires matching preparation provenance before reuse; the reviewer rechecked the fix and is clear. All required local checks pass, including 3,873 unit tests and 78 end-to-end tests.

janicduplessis:
Expanded the untimed cache verification after review: all 366 compilation units in the current native compiler databases directly hit the original golden cache under the pinned locale, across the app and installed native libraries. Probes used read-only cache access and restored the original object/dependency files. This does not replace or change the failed timed run.

PR #662, opened 2026-09-10
chore: remove the experimental Android CAS compiler cache

### LINKED ISSUE #658: Remove the experimental Android CAS compiler cache
## Why

`optimizations.android.compilerCache="cas"` and `optimizations.android.casToolchain` select an experimental Apple Clang toolchain for Android native compilation. The setting is a live footgun:

- It is the only optimization that depends on a path outside the repository, so it rots. A machine here had `casToolchain` pointing into a stim worktree that had been deleted. Every `stim android` then died instantly with `STIM_BAD_ARG: Could not configure Android build: ENOENT ... toolchain.json`, and `stim doctor --platform android` reported clean.
- Clearing the dead path is not enough. Removing only `casToolchain` leaves `compilerCache: "cas"`, and the next run refuses with `optimizations.android.compilerCache=cas requires optimizations.android.casToolchain or STIM_ANDROID_CAS_TOOLCHAIN`. Two refusals to reach a working build.
- `casToolchain` is not in the machine-settings section of `stim guide settings` where a reader would look for it, so recovering means guessing.

It is experimental, it needs a manifest we do not ship, and ccache covers the same ground without an external dependency.

## What to remove

- `'cas'` from the `optimizations.android.compilerCache` choices, leaving `auto | ccache | none`. `auto` becomes simply ccache.
- `optimizations.android.casToolchain` and the `STIM_ANDROID_CAS_TOOLCHAIN` environment variable.
- `packages/stim-cli/src/engine/android-cas.ts` and its call in `commands/android.ts`.
- The `cas` parameter threaded through `engine/gradle.ts`, and the `STIM_ANDROID_CCACHE` branch that exists only to turn ccache off for CAS.
- The CAS paragraphs in `guide settings` and `guide lifecycle`, and the `android-cas` docs link.

## Degrade, do not refuse

This is the part to get right. A machine that already has `compilerCache: "cas"` set must keep working: read it, ignore it, use ccache, and say so once in a dim line naming the setting and the file. Do not throw `Invalid optimizations.android.compilerCache`, and do not make the user edit a config file before they can build. The same goes for a leftover `casToolchain` key and a set `STIM_ANDROID_CAS_TOOLCHAIN`: both become inert, neither is an error.

`stim doctor` reports the retired keys as a `note` with the line that clears them, so the config gets tidied on someone's own schedule rather than at the moment they wanted a build.

## Tests

- A settings file with `compilerCache: "cas"` and a valid `casToolchain` builds with ccache and emits the retirement note.
- The same with a `casToolchain` pointing at a path that does not exist: still builds, still one note, no refusal. This is the case that broke a real machine.
- `STIM_ANDROID_CAS_TOOLCHAIN` set in the environment changes nothing.
- `compilerCache: "auto"` and an absent setting both select ccache.
- Guide contract test: no topic mentions `cas`, `casToolchain`, or the env var.

## Not in scope

ccache behavior, `optimizations.android.pch`, and the iOS compilation cache, which is a separate mechanism and stays.

janicduplessis:
## Description

`optimizations.android.compilerCache="cas"` and `optimizations.android.casToolchain`
selected an experimental Apple Clang toolchain for Android native compilation.
It was the only optimization that depended on a path outside the repository, so
it rotted: a machine here had `casToolchain` pointing into a deleted worktree,
every `stim android` died with `STIM_BAD_ARG: Could not configure Android build:
ENOENT ... toolchain.json`, and `stim doctor --platform android` reported clean.
Clearing the dead path alone was not enough, because `compilerCache: "cas"`
without a manifest refused with its own error.

The backend was experimental, needed a manifest Stim does not ship, and ccache
covers the same ground with no external dependency.

Blast radius: anyone who had CAS configured now compiles under a different
native profile directory (`.cxx/stim-<profile>`), so their next Android build is
a cold C++ build once. Everyone else is unaffected -- the ccache, `none`, and
`project` backends keep the same `STIM_ANDROID_NATIVE_PROFILE` inputs they had.

## Solution

Removed the backend: `engine/android-cas.ts`, `bin/android-cas-compiler.ts` and
its tsdown entry, the three `shim/android-cas.*` files, `scripts/android-cas/`,
the `cas` plumbing in `commands/android.ts` and `engine/gradle.ts` (including the
`STIM_ANDROID_CCACHE` branch that existed only to turn ccache off for CAS), and
the CAS paragraphs in `guide settings`, `guide lifecycle`, the website, and the
package README. `website/docs/android-cas.md` and `docs/android-cas-poc.md` are
deleted with the sidebar entry.

The part that mattered is the degradation, not the deletion. A machine that
still has the old settings keeps building:

- `compilerCache: "cas"` resolves to ccache instead of refusing. It is not a
  value error, so nothing throws `Invalid optimizations.android.compilerCache`.
- `casToolchain` is dropped from `OPTIMIZATION_SHAPES` and skipped by
  `unknownSettingKeys`, so it is inert rather than either a shape error or a
  yellow "not read by Stim" warning. Its value is never validated, so a path
  into a deleted worktree cannot fail a build.
- `STIM_ANDROID_CAS_TOOLCHAIN` is no longer read anywhere. `resolveOptimizations`
  no longer takes an `env` argument at all.

`stim android` says so once, in a dim `setting` line naming the keys and the
file they came from, then builds:

```
  setting     Ignoring retired optimizations.android.compilerCache, optimizations.android.casToolchain in /Users/j/.stim/config.json. Android builds use ccache.
```

`stim doctor` reports the same keys as a `note` with the line that clears them,
so the config gets tidied on someone's own schedule rather than at the moment
they wanted a build:

```
note  A retired Android compiler cache setting is still configured
  Ignoring retired optimizations.android.compilerCache, optimizations.android.casToolchain in /Users/j/.stim/config.json. Android builds use ccache.
  -> Delete optimizations.android.compilerCache and optimizations.android.casToolchain from /Users/j/.stim/config.json.
```

Naming the file needs provenance the merged settings do not carry, so
`resolveSettings` now builds its layers through a `settingsLayers` helper that
carries each layer's file, and `retiredSettings` reports the highest-precedence
layer that actually supplies each retired key. A key that a live higher layer
overrides is not reported, because it is not what the build ignores.

Deliberate trade-offs:

- The remedy is an editing instruction, not a runnable command. Stim has no
  config command (and should not grow one), and a one-liner that rewrites
  `config.json` would do an unlocked read-modify-write on the file that holds
  the device records.
- `androidBuildOptions`'s `compiler` option is removed, since CAS was the only
  thing that ever set it. The generic `compiler` cache-key dimension in
  `@stim-cli/core` stays: it is part of the published cache-key contract the
  cache packages implement, and `build-cache.test.ts` still pins CLI/provider
  agreement on it.
- A machine that used CAS keeps a `~/.stim/android-cas/<id>` directory and its
  `caches.json` entry. The manifest is generic and `registeredCaches` filters on
  directory existence, so `stim gc` can still name and reap it; nothing is
  deleted automatically.

## Test plan

From the repo root on this branch, rebased onto `dd01dcad6`:

- `pnpm format:check` -- clean
- `pnpm lint` -- clean
- `pnpm typecheck` -- clean
- `pnpm knip` -- clean
- `pnpm test` -- 121 files, 3867 passed, 1 skipped

New and changed coverage:

- `android-command.test.ts` -- a `.stim.json` with `compilerCache: "cas"` plus a
  `casToolchain` both for a manifest that exists and for one whose worktree is
  gone: the build runs with the ccache setup, emits exactly one retirement note
  naming both keys and the file, emits no "not read by Stim" warning, and
  refuses nothing. This is the case that broke a real machine.
- `android-command.test.ts` -- `STIM_ANDROID_CAS_TOOLCHAIN` set in the
  environment changes nothing and produces no note.
- `optimizations.test.ts` -- `"cas"` with and without a toolchain, a bare
  `casToolchain`, `"auto"`, and an absent setting all resolve to ccache;
  `"none"` still resolves to `none`; a genuinely unknown value still refuses.
- `guide.test.ts` -- contract test that no rendered topic mentions
  `casToolchain`, `STIM_ANDROID_CAS_TOOLCHAIN`, `android-cas`, `Android CAS`, or
  `compilerCache ... cas`.

Real-command verification (invariant 9), against the built `dist/cli.mjs` on a
throwaway project with a temporary `STIM_HOME`:

- machine `config.json` carrying `compilerCache: "cas"` and a `casToolchain`
  pointing at a deleted worktree, with `STIM_ANDROID_CAS_TOOLCHAIN` also set:
  `stim android` printed the dim `setting` line above and then continued to the
  normal `STIM_NO_METRO` check -- it did not refuse on the setting.
- the same keys in a project `.stim.json`: `stim doctor --platform android`
  printed the `note` above, where it previously reported clean.

Closes #658


janicduplessis:
Reviewed with Codex (gpt-6-astra) in addition to the fresh-agent pass. One finding, reproduced and fixed in `68e7ca6`-onwards: with `compilerCache: "none"` and a leftover `casToolchain`, the retirement note claimed `Android builds use ccache` while the build correctly honoured `none` and disabled it.

The note now asserts a backend only when it knows one. A retired `compilerCache` means ccache genuinely is the fallback, so it says so; when only the toolchain path is retired, `compilerCache` still decides and the note stays silent about the backend.

Verified by hand against the exact config that broke a real machine, plus the `none`, `ccache`, and `auto` values, an invalid value (still rejected, as a `costs time` finding), and the `STIM_ANDROID_CAS_TOOLCHAIN` env form (inert). No CAS files remain in the package; the only surviving references are the three constants that let the retired value be recognised and ignored.

PR #664, opened 2026-09-10
fix: retain harmless benchmark session refusals

### LINKED ISSUE #663: Avoid rejecting benchmark read-only attempts that never reached a device
## Problem

The benchmark isolation audit rejects an otherwise complete run when an unscoped read-only agent-device command fails because no session exists. It reports `agent-device-run-session-not-applied` even though the command never reached a device and all successful interactions use the run session.

## Evidence

An Android control run attempted `agent-device screenshot /tmp/initial.png`; the captured tool output was `Error (SESSION_NOT_FOUND): No active session. Run open first.` The agent then retried with the prescribed state directory and session. Its Settings screenshot, recording and device proof passed; only the session audit rejected it. The failed command is retained in the transcript and elapsed time.

## Cause

`agentDeviceIsolationInvalidReasons` checks command strings without considering a confirmed no-session refusal. Its broad rule correctly rejects unscoped device use, but also treats this failed read-only attempt as device use.

## Fix idea

Allow only literal read-only screenshot/snapshot attempts whose captured result proves SESSION_NOT_FOUND, without an explicit device/session override. Keep successful or ambiguous unscoped operations rejected, preserve the attempt and timing, and rerun the audit against retained evidence rather than rerunning the agent.

janicduplessis:
## Description

A complete Android control benchmark was rejected because one bare screenshot attempt returned `SESSION_NOT_FOUND`. It never reached a device; every successful interaction used the required run session.

## Solution

Keep the [per-command isolation policy](https://github.com/appandflow/stim/commit/782b8c5046cf300fb54dddf509ca9240cdb61cf5), but exempt literal screenshot/snapshot attempts with a captured no-session refusal. Successful unscoped commands, explicit target overrides and ambiguous command chains still fail. The failed attempt and its elapsed time remain in the result.

## Test plan

Replayed all 54 retained commands from the affected run: the session audit now passes without modifying its transcript. Settings and recording proof remain subject to the existing checks. Regression cases reject successful reads, overrides, hidden commands and output that indicates a device was reached.

Fixes #663.

### DIFF
--- a/scripts/agent-benchmark/README.md
+++ b/scripts/agent-benchmark/README.md
@@ -287,6 +287,11 @@ events and reruns the current session audit. It can clear only the sole
 `agent-device-run-session-not-applied` reason; every evidence check still applies.
 Keep the original verdict and correction provenance private, outside Git.
 
+Literal unscoped screenshot/snapshot attempts that report `SESSION_NOT_FOUND`
+without reaching a device remain visible in timing but do not fail isolation.
+Successful unscoped reads, explicit device/session overrides and ambiguous
+multi-command results still fail the audit.
+
 For launch-error runs, `launch-error-audit-review.json` records a review of each
 unrecognized setup command. It contains `schemaVersion: 1`, `runId`,
 `originalRecordSha256`, `metaSha256`, `policySha256` (the launch-crash audit source),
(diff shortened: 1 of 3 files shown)


janicduplessis:
Fresh correctness review is clear after inspecting shell ambiguity cases, policy tests and guidance. All 42 run-guards tests and the full required local checks pass. Replaying the retained 54-command run clears only the no-session false positive; original commands and timing remain unchanged.

PR #666, opened 2026-09-10
fix: recognize persisted benchmark crash evidence

### LINKED ISSUE #665: Recognize crash logs read from Claude persisted tool output
## Problem
A completed Sonnet Android launch-crash control run is rejected as launch-crash-error-capture-missing, although the agent read the captured runtime error and repaired the app.

## Evidence
Run sonnet-android-launch-error-sonnet-launch-crash-control-1789072172083. adb logcat tool call toolu_012Pnj8CbSFBJBpe7myceGBF returns a persisted-output envelope (36.3KB). Follow-up toolu_01Q9j9GpJdwVBZVahNvbEPpS searches that exact tool-results/bilbem0m6.txt and returns the injected error and RootLayout frame before source inspection. Settings screenshot, repair and recording pass.

## Cause
scripts/launch-crash-benchmark.mjs errorCaptureCommand recognizes direct device log commands and .log/tmp paths, but not the linked .txt path in a prior captured-output envelope. Confirmed still present on origin/main.

## Fix idea
Follow only exact persisted output references from prior successful runtime log commands. Require the later captured output to contain the token and retain its later timestamp. Do not accept arbitrary text/source files or fabricate earlier evidence; re-audit original run without agent retry.

janicduplessis:
## Description

Sonnet repaired the Android launch crash and reached Settings, but the audit missed its evidence: Claude saved long log output to a `.txt` file, and the initial Expo command used a wrapper the launch parser did not recognize. Fixes #665.

## Solution

Follow exact saved-output references from runtime log commands, timing diagnosis when the agent reads the error. For the wrapped Expo pipeline, require a later read of its exact build log to show both build success and app opening; an echo-masked exit alone is insufficient. The pipeline-success safeguard from [#515](https://github.com/appandflow/stim/commit/ca7df6cd2dfccd453171cf59d03197e5de84f6bc) remains.

This changes benchmark analysis only, not Stim or retained transcripts.

## Test plan

Replayed the retained Sonnet transcript: diagnosis resolves to the saved-log read before source inspection and repair. Regression cases reject unrelated files, failed reads, overlapping capture/read times and pipelines without observed launch evidence.


janicduplessis:
Fresh correctness review caught unrelated-file substitution in downstream filters. Fixed by restricting the initial reader to the sole referenced file and downstream commands to stdin-only filters, with four negative regression cases. Fresh re-review is CLEAR; all defined local checks rerun and pass (3870 tests, 78 e2e). Original retained transcript still diagnoses at the later saved-log read; no run data or published runtime changed.

PR #670, opened 2026-09-10
fix: retain cache evidence from scoped benchmark commands

### LINKED ISSUE #669: Accept scoped status wrappers for retained compiler-cache evidence
## Problem
Opus Android native Stim1789075408944 is rejected with ccache-evidence-missing despite a timestamped386/386-hit stats file and successful27.6s build.

## Evidence
The build command is cd <exact worktree> && stim android --system-image <image>; echo "EXIT=$?". Output reports build ok then adopted pm clear failure, EXIT=1. Retry uses artifact cache and reaches Settings. Stats mtime21:26:58.842Z lies within original build command21:26:28.897Z–21:27:07.277Z.

## Cause
ccacheLogEvidence requires exactly one shell segment, so the scoped cd and explicit status-report wrapper prevent retaining compiler evidence. Runtime print omission is separately tracked in #644; this issue concerns benchmark validation.

## Fix idea
Accept a narrow literal cd to the recorded worktree plus explicit status echo, preserving exact command/event/meta hashes, unique build and timestamp/ownership checks. Reject unrelated commands, wrong worktrees and ambiguous/mismatched status reports. Re-audit retained run without repeating agent.

janicduplessis:
## Description

Opus's native benchmark hit the compiler cache 386/386 times, but the audit discarded the evidence because its command wrapped Stim in `cd ... && stim android ...; echo "EXIT=$?"`. Fixes #669.

## Solution

Accept this narrow wrapper only for the recorded worktree with one explicit status report. Preserve the command/event hashes, unique-build requirement and stats-log timestamp checks introduced in [#646](https://github.com/appandflow/stim/commit/2acde6e53433e9499e8b723f37df3c025756eb62). Unrelated commands and directories still refuse. This changes benchmark analysis only; the device-cleanup retry remains in the timing.

## Test plan

Replay the retained Opus build: recover 386 hits and zero misses without editing commands. Regression cases reject wrong directories, unconditional `cd`, extra commands, substitutions and missing/duplicate/invalid status reports.

### DIFF
--- a/scripts/agent-benchmark/README.md
+++ b/scripts/agent-benchmark/README.md
@@ -68,6 +68,10 @@ crashed command that could have compiled still fails closed.
 Completed tool output triggers an immediate `CACHE ALERT` and preserves
 `cache-alerts.json`. A flagged attempt stays available for investigation and is
 excluded from published comparisons; investigate the cause before retrying.
+When a completed build fails later during launch, collection can retain the
+owned workspace's timestamped compiler statistics. A literal `cd` to that
+worktree followed by `&& stim android ...; echo "EXIT=$?"` is also eligible;
+additional commands, a different directory, or ambiguous status reports are not.
 
 Android golden preparation and every preflight require a structured doctor
 report without cost findings. Repair the fixture with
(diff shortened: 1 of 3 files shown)


janicduplessis:
Fresh correctness review CLEAR. Independent reviewer ran both relevant suites (67 tests). Full local checks passed (3878 tests plus one skip, 78 e2e, format/lint/build/typecheck/runtime/knip). Retained Opus evidence replay recovers386/386 hits with unchanged commands and in-window stats mtime.

PR #679, opened 2026-09-11
feat: validate the machine config, and degrade instead of refusing

### LINKED ISSUE #658: Remove the experimental Android CAS compiler cache

### LINKED ISSUE #660: doctor should validate the machine config, not leave commands to fail on it
## Why

`~/.stim/config.json` can hold settings that are individually well-formed and jointly useless. Nothing checks it until a command reads it, and the refusal names the missing *thing* rather than the *setting* that pointed at it.

A real sequence from this machine:

    stim android
      STIM_BAD_ARG: Could not configure Android build: ENOENT ...
      /Users/j/Developer/stim/.worktrees/pch-portability-.../toolchain.json

    stim doctor --platform android
      (clean)

The toolchain path was a stim worktree deleted weeks earlier. Doctor -- the command whose job is "your machine is not set up for fast builds" -- had nothing to say, and the build error named a file the user never typed, not `optimizations.android.casToolchain`, the key that named it.

`doctor` already reads the settings this file holds. It should check them.

## What to report

All at `note`, each naming the key, the file, and the line that clears it:

- **A setting whose value is a path that no longer exists.** Generic over the settings tree: any key declared as a path type in `optimizations.ts` (`'path'`) plus the other path-valued settings, checked with a plain existence test.
- **A setting that requires a companion the config does not supply**, rather than waiting for the command to refuse. These pairs are already encoded as throws in `optimizations.ts`; doctor should surface them as findings instead of leaving them to fire mid-build.
- **A key Stim no longer reads.** A retired setting is invisible today: it sits in the file looking meaningful. Name it and say it is inert.
- **A file that is not valid JSON**, with the repair line `config.ts` already prints for `STIM_CONFIG_CORRUPT`.

Skip the `projects` subtree. An entry for a checkout that no longer exists is normal there, and `gc` owns it -- reporting those would bury the settings findings in noise. (Verified: a plain existence sweep over the whole file on this machine produced one finding, and it was a dead `projects` entry.)

## Where

`doctor.ts`, as its own check beside the existing ones, so it runs on every `stim doctor` regardless of `--platform`: a bad machine setting is not a native-platform concern. Pure decision logic separate from I/O, matching how the file is already organised, so the "which settings are broken" decision is unit-testable against a settings object without touching disk.

## Tests

- A path setting pointing at a missing file produces one finding naming the key.
- The same path existing produces none.
- A `projects` entry pointing at a deleted checkout produces none.
- An unparseable config produces the corrupt-file finding, not a crash.
- A retired key produces the inert-key finding.
- A config with nothing wrong produces no findings.

## Related

appandflow/stim#658 removes the specific setting that motivated this. This issue is the general check, and stands on its own: the next path-valued setting will rot the same way.

janicduplessis:
## Description

`~/.stim/config.json` can hold settings that are individually well-formed and
jointly useless, and nothing checked them until a command read one. A machine
here had `optimizations.android.casToolchain` pointing into a stim worktree
deleted weeks earlier. Every `stim android` died instantly with
`STIM_BAD_ARG: Could not configure Android build: ENOENT ... toolchain.json`,
naming a file the user never typed, while `stim doctor --platform android`
reported clean. Clearing the dead path was not enough: the next run refused with
`optimizations.android.compilerCache=cas requires
optimizations.android.casToolchain or STIM_ANDROID_CAS_TOOLCHAIN`. Two refusals
to reach a working build.

Two causes, one diff because they meet in `doctor.ts`:

- `resolveOptimizations` threw for an unusable CAS toolchain and for a `cas`
  selection with no toolchain, so a rotted path outside the repository could
  block every Android build. Separately, `resolveAndroidCas` validated only part
  of the manifest, so a manifest it accepted could still produce a compiler that
  could not compile -- and that failure lands after the point where `android.ts`
  can degrade.
- `doctor` read the settings this file holds but checked none of them, and a
  corrupt `config.json` crashed it at `getConcurrencyLimits` rather than
  reporting the file.

CAS itself stays. #671 reverted an earlier attempt that removed the feature.

### What this branch caused, and what it inherited

Setting `casToolchain` to `null` -- the most natural way to clear it -- also
refuses from the shape gate. Measured, with
`{ "compilerCache": "none", "casToolchain": null }` in `~/.stim/config.json`:

| build | `android` | `ios` | `gc` |
|---|---|---|---|
| published `1.0.0` | ok | ok | **refuses** |
| `main` (measured at `99de2b229`) | ok | ok | **refuses** |
| this branch, first commit only | **refuses**, all four `compilerCache` states | -- | refuses |
| this branch now | ok | ok | ok |

Only the `gc` refusal is pre-existing. `settingShapeErrors`' call sites are
byte-identical between `main` and this branch, so the gate was not widened; what
moved is the declaration. `main` declares
`'optimizations.android.casToolchain': 'path'` only in `optimizations.ts` and
never in `settings.ts`, so the shape table `settingShapeErrors` reads did not
carry the key. Adding the doctor sweep brought it into that table, which made
`android`, `ios`, `start` and `worktree warm` refuse a non-string value too.
**This branch introduced that breadth in its first commit and then fixed it**, which
is also why `SETTINGS_WITH_RESOLVE_TIME_FALLBACK` is the right shape for the
fix: the key has to be visible to the sweep and exempt from the gate.

## Solution

**Degrade, do not refuse (#658).** `resolveOptimizations` no longer throws for a
toolchain it cannot use. It returns `android.compilerCacheFallback` describing
the setting at fault and leaves `compilerCache` at the backend that will
actually run, then `stim android` prints one warning and builds.
`optimizations.android.casToolchain` is also exempt from the shape gate
(`SETTINGS_WITH_RESOLVE_TIME_FALLBACK`), because that gate refuses before the
fallback can run: any value it cannot use -- `null`, a number, a boolean, an
object, a relative path, a trailing newline -- now degrades instead of refusing
`android`, `ios`, `start`, `worktree warm`, and `gc`. A value outside
`auto|ccache|cas|none` still refuses by name: that is a typo, not a rotted path.

The toolchain read itself can fail, so the read and the degrade it produces live
in `engine/android-cas.ts` rather than in the command: `readAndroidCasToolchain`
owns the read, and `resolveAndroidCompilerCache` runs an attempt and degrades
`cas` to `ccache` with a fallback naming the key, the reason, and the manifest it
tried. `commands/android.ts` passes the full setup as that attempt;
`doctor.ts` and `doctor-cxx.ts` pass the read alone. One degrade, one warning
formatter, so the build and the report cannot disagree about the same config.
For that degrade to be the thing that runs, `readAndroidCasToolchain` has to
reject every manifest that cannot produce a working compile, not only the ones it
cannot parse:

- all **seven** manifest fields must be strings. `resourceDir` was not checked,
  and the compiler shim passes it to clang as `-resource-dir`, so a manifest
  without it passed setup and then failed the real compile with
  `fatal error: 'stddef.h' file not found`.
- `clang`, `clangxx`, `lld`, `ar` and `ranlib` must be **executable regular
  files**, not merely present. A readable non-executable compiler passed setup
  and then failed with `spawn ... EACCES`.
- `resourceDir` must be an existing directory.

The fallback reason now reads `could not be used:` rather than `could not be
read:`, because a manifest that parses and names an unusable toolchain was read
fine.

**What the manifest validation does not cover.** It does not check that
`resourceDir` actually holds clang's builtin headers, and it does not check the
two paths inside `ndk` that the shims dereference:
`build/cmake/android.toolchain.cmake`, included by the CAS CMake toolchain file,
and `toolchains/llvm/prebuilt/darwin-x86_64/sysroot/usr/include/c++/v1`, which
the compiler shim passes as `-isystem` for every C++ translation unit. A
partially deleted NDK whose `source.properties` survives therefore still fails
during the build rather than degrading. Those are internals of a field the
manifest already declares, not manifest fields, and the `darwin-x86_64` segment
is a hardcoded host assumption in the shim; covering them is a follow-up.

The warning names the backend the selection leaves, and words it so the next
line cannot contradict it. `ccache` is not guaranteed to run -- `resolveCcache`
returns `null` when ccache is not installed and when the project's
`build.gradle` declares its own `CMAKE_*_COMPILER_LAUNCHER` -- so the sentence
is "Android builds fall back to ccache when it is available", which the
following `ccache on|off` line completes rather than falsifies. With
`compilerCache: "none"` the line says no compiler cache runs, and why.
`compilerCacheFallbackMessage` is the single formatter for that sentence, shared
by the build warning and the doctor finding, so the two cannot drift.

**doctor validates the config (#660).** `doctor-config.ts` holds a pure
`checkMachineSettings` decision (injectable `exists` and `env`, no disk needed)
beside a thin `readMachineSettings` that owns the I/O, matching how `doctor.ts`
and `sandbox.ts` are already split. It runs on every `stim doctor` regardless of
`--platform`, and reports at `note`, each naming the key, the file it came from,
and the line that clears it:

- a path setting whose value no longer exists, generic over the `'path'` shapes
  already declared in `optimizations.ts` and `settings.ts` (`casToolchain`,
  `ios.simslimProfile`, `android.avdConfigFile`, `android.keystore`), relative
  values resolved against the project root;
- a setting, or an environment variable, holding a value Stim cannot use, and a
  setting that needs a companion the config does not supply, surfaced from the
  fallback above instead of firing mid-build;
- a key Stim no longer reads, via the existing `unknownSettingKeys`;
- a config that is not valid JSON, with the repair line `config.ts` already
  prints for `STIM_CONFIG_CORRUPT`, now factored out as `configCorruptRepair`
  so the two cannot diverge.

Deliberate choices a reviewer should weigh:

- **The environment is a source in the existence sweep, and it wins.** With no
  configured `casToolchain` and `STIM_ANDROID_CAS_TOOLCHAIN` naming a file that
  is not there, resolution produces no fallback -- the value is a clean absolute
  path -- so the sweep found nothing while `stim android` warned. The sweep now
  reads `STIM_ANDROID_CAS_TOOLCHAIN` for that key and attributes the finding to
  the environment. Precedence matches `resolveOptimizations` and matches how
  `settingOrigin` already reports only the winning layer: when the variable
  holds a usable path, a dead configured value is not reported, because nothing
  reads it. The one-finding rule survives, because the sweep skips the key when
  the fallback named either the key or its variable.
- **The sweep tests the string the build opens, per setting.** For
  `casToolchain` that is the configured value verbatim, which is what
  `resolveAndroidCas` passes to `readFileSync`: a value whose path ends in a
  space is reported missing only when it really is, and is shown with
  `JSON.stringify` when it differs from its trimmed form, so invisible
  whitespace is visible in the finding. `android.keystore` is trimmed, because
  `resolveKeystore` trims before opening it -- swept verbatim, a keystore of
  `" release.keystore "` beside an existing `release.keystore` built fine and
  was still reported as a path that is not there. `ios.simslimProfile` and
  `android.avdConfigFile` stay verbatim: their consumers refuse an untrimmed
  value rather than accepting it, so an untrimmed value there is invalid, not
  missing, and is reported as such.
- **doctor resolves the manifest, and so do its downstream gates.** `doctor`
  degrades through `resolveAndroidCompilerCache` before it reports anything, so
  the value every later check reads is the one the build will use. Three measured
  defects close together: a manifest that exists but cannot be used is a finding
  in the build's own words rather than silence; `androidCcacheFindings` runs on a
  machine whose `cas` selection degraded, so "ccache is not on PATH" is reported
  to the machine that now needs ccache; and `repairCxxLauncherState` no longer
  short-circuits there, so `stim doctor --fix` removes the launcherless legacy
  `.cxx` that machine compiles against. The **cost** is that a read-only report
  now does filesystem work it did not do before: one `readFileSync` of the
  manifest, one of the NDK's `source.properties`, and a `stat` of the five
  compiler binaries plus `resourceDir`. It spawns nothing, and it runs only when
  `compilerCache` resolves to `cas`. That is a fraction of what `doctor` already
  does on every run -- it walks `node_modules` for `.cxx` CMake caches, reads
  each one, and spawns `git ls-files` and `git check-ignore` per candidate -- and
  a report that cannot see what the next build will do is the defect this closes,
  so the read is worth its cost. The gating alone, without the read, would have
  left the silent unusable manifest open.
- **A path that is not there keeps its own finding.** The read degrades for a
  missing manifest too -- that is how `--fix` becomes eligible on the motivating
  machine -- but the existence sweep, not the fallback, reports it, because
  "names X, which does not exist" and "point it at the path it should name" beat
  relaying `ENOENT` from the resolver. The fallback finding yields whenever the
  manifest it names is absent, so this stays one finding per key, and the only
  new output is a manifest that is present and unusable.
- **The `projects` subtree is skipped.** An entry for a deleted checkout is
  normal there and `gc` owns it. Driving the sweep from the declared `'path'`
  shapes excludes it structurally rather than by an exception. Verified on this
  machine: a plain existence sweep over the whole file produced exactly one
  finding, a dead `projects` entry.
- **A corrupt config returns that finding alone.** Every other machine-level
  check reads the same file and would throw. Reporting the unreadable file and
  stopping is the honest result, and it is what stops doctor crashing --
  including under `--fix`, where `repairCxxLauncherState` read the same file
  through `projectOptimizations`. A config that cannot be read now repairs
  nothing, rather than printing a raw parse error and exiting 1.
- **One key, one finding.** The path sweep skips any key the fallback already
  named, so a relative value or a trailing newline is reported once, by the
  check that knows the key is absolute-only, instead of twice with a
  `<projectRoot>/...` path `resolveOptimizations` never resolves. It also skips
  `ios.simslimProfile` when `checkSimSlim` already reported it in the same run.
- **A config whose optimizations cannot be resolved skips the keys that resolve
  owns.** A typo in `compilerCache` -- `"CAS"` -- makes `resolveOptimizations`
  throw, so there is no fallback to suppress the sweep, and the sweep then fired
  unconditionally on `casToolchain`: a second finding naming an invented
  `<projectRoot>/tc/toolchain.json` for a relative value, or a filesystem path
  carrying a raw newline. When `optimizations` is null the sweep now skips
  `SETTINGS_WITH_RESOLVE_TIME_FALLBACK`, the same declaration the shape gate
  already exempts. The typo is reported by name, once; fixing it reveals whatever
  the toolchain key is doing next run.
- **A dead path finding claims only what doctor knows.** It names the key, the
  file and the path, and stops. Whether a command reads that value depends on
  `compilerCache` and on the environment, so the finding does not assert that
  one does.
- **Project- and repo-scoped settings name their subtree.** Both live in
  `config.json`, so the finding says
  `config.json (projects["<root>"].settings)` rather than sending the reader to
  the top of the file.
- **`stim doctor --fix` gains eligibility on a machine whose CAS setting
  rotted.** After a resolve-time fallback `compilerCache` is `ccache`, so
  `repairCxxLauncherState` no longer short-circuits, and a checkout's ignored,
  untracked *legacy* `.cxx` configurations become removable -- which is correct,
  because that machine now compiles with ccache. CAS build output is not
  affected: Stim's own staging lives at `.cxx/stim-<profile>`, which
  `readCxxLauncherStates` excludes at depth 0, and a test asserts the managed
  profile survives while the legacy directory is removed. This now covers a dead
  *absolute* path, the motivating case, because the repair resolves the manifest
  the way the build does; a toolchain Stim can use still keeps `cas` selected and
  still repairs nothing.
- **`tempDir` and `caches.buildCache` are not covered.** They are top-level
  `Config` keys, not declared settings, so including them would mean the
  hardcoded list #660 rules out. Worth a follow-up issue.

### How this sits on the work merged under it

Rebased onto `main` after #681 (the ownership-claim primitive), #690
(`worktree warm --refresh`), #661 (the two named workflows and the gated seed
findings) and #700 (the "main checkout" -> "source checkout" rename in shipped
text). Three resolutions a reviewer should know about:

- The `doctor --help` phase line keeps #700's wording and gains this branch's
  scope: `settings  every Stim setting type, machine config paths, companions
  and inert keys`.
- The machine-config checks sit **outside** #661's linked-worktree gate. That
  gate covers the seed findings, which are a worktree-workflow concern; a rotted
  machine setting is not, so `checkMachineSettings` runs on every `stim doctor`
  whatever `--platform` says. Verified by running: a repository with no linked
  worktree reports no source-checkout seed finding and still reports a dead
  machine path.
- #690's remedy for the behind-finding is untouched: it still names
  `stim worktree warm --refresh`, confirmed from the built CLI against a
  checkout one commit behind its upstream.

Shipped text added by this branch uses "source checkout". The three surviving
"main working tree" / `/w/main checkout` strings in the repository are git's own
concept (how `git worktree list` orders output, git's refusal to remove a
repository's main working tree, and a fixture command string), not Stim's term
for the role, and are unchanged.

## Test plan

From the repo root, outside the shell sandbox:

- `pnpm format:check`, `pnpm lint`, `pnpm build`, `pnpm typecheck`, `pnpm knip` -- clean.
- `pnpm test` -- 130 files, 4064 passed, 1 skipped.
- `pnpm test:e2e` -- 79 passed (this changes `doctor` and `android` behavior).

New coverage, by the case each catches:

- `android-cas.test.ts`: the compile boundary, not just setup. A complete
  manifest's setup is driven end to end -- `resolveAndroidCas`, then the `clang`
  wrapper it symlinks, then the real `dist/android-cas-compiler.mjs`, then the
  compiler it spawns -- and `compiler.jsonl` shows `-resource-dir <resourceDir>`
  and exit 0. The same compiler run against a context whose `resourceDir` does
  not exist exits non-zero with `'stddef.h' file not found`, and setup now
  refuses both a manifest with no `resourceDir` and one naming a `resourceDir`
  that is not a directory. A compiler that is readable but not executable fails
  `runFile` with `EACCES` and is refused at setup naming the fields. A manifest
  that parses but names no compiler throws naming every field it lacks.
- `android-command.test.ts`: the run falls back rather than failing. A manifest
  with no `resourceDir`, and one whose compiler is not executable, each build in
  full with `{ cas: null, ccache: <setup> }`, no `apple-cas-` cache key, and one
  warning asserted exactly. Also: a non-string `casToolchain` builds in all four
  `compilerCache` states, asserting the exact warning and the engine options;
  the warning does not promise ccache when `ccacheFor` returns `null`; `cas`
  with no toolchain falls back; `compilerCache: "none"` with a dead toolchain
  key stays inert and the output never mentions ccache; a non-absolute path with
  `none` reports the `none` backend; an unusable manifest in the environment
  names the environment.
- `doctor-config.test.ts`: an absent environment-sourced toolchain is one
  finding titled against the environment with the `unset it` remedy; an
  environment toolchain that exists silences the configured value nothing reads;
  a value whose path ends in a space is absent only when it really is absent, and
  is quoted in the detail when it differs from its trimmed form; an
  `android.keystore` whose value carries surrounding spaces is absent only when
  the path its consumer opens is absent, and the finding names that path; `runDoctor`
  reports the environment-sourced dead path end to end. Plus, unchanged: the
  non-string cross product produces exactly one finding per `compilerCache`
  state; a relative value and a trailing newline each produce one finding that
  invents no project-root path; a path setting naming a missing file produces one
  finding asserting only what doctor knows; the same path existing produces none;
  a relative `ios.simslimProfile` resolves against the project root; a `cas`
  selection with no toolchain is a finding, not a mid-build refusal; an inert key
  is named; a clean config produces none. On disk: a config carrying both a dead
  `projects` entry and a real project-scoped `android.keystore` reports the
  keystore -- naming the `projects[...].settings` subtree -- and says nothing
  about the dead checkout; a setting committed in `.stim.json` is reported
  against that file; an unparseable config produces the corrupt finding and
  `runDoctor` returns only it; a dead machine path is reported under
  `--platform ios`, `--platform android`, and neither. New: a manifest that is
  there and declares no fields is one finding whose detail is the exact sentence
  `stim android` warns with and whose remedy names the manifest to repair; the
  ccache checks are reported for a dead toolchain and not for a usable one, which
  is the gate following the degraded backend; and a `compilerCache` of `"CAS"` is
  exactly one finding that names the key, with no `<projectRoot>/tc` path -- at
  the `checkMachineSettings` boundary for a relative value and for one carrying a
  raw newline, and through `runDoctor` on disk.
- `optimizations.test.ts`: the full cross product of non-string `casToolchain`
  values (`null`, a number, a boolean, an object, an array) against all four
  `compilerCache` states -- no row throws, no row produces a shape refusal, and
  every row's message names the backend that row resolved.
- `doctor-cxx.test.ts`: a CAS selection Stim cannot use -- a relative value, and
  the dead absolute path of the motivating machine -- removes the legacy `.cxx`
  configuration and leaves `.cxx/stim-<profile>` in place, while a toolchain Stim
  can use keeps both, from the setting and from the environment; a config that
  cannot be read repairs nothing. The usable-manifest fixture is
  `writeCasToolchain` in `_factories.ts`, which replaces the copy
  `android-command.test.ts` held.

Real-command evidence, built CLI against a temporary `STIM_HOME`:

```
$ STIM_ANDROID_CAS_TOOLCHAIN=/missing/toolchain.json stim doctor --platform android
note  An environment variable points at a path that is not there
  STIM_ANDROID_CAS_TOOLCHAIN in the environment names /missing/toolchain.json,
  which does not exist.
  -> Point STIM_ANDROID_CAS_TOOLCHAIN at the path it should name, or unset it.

$ stim doctor --platform android          # casToolchain: "<home>/toolchain.json " (trailing space)
note  A setting points at a path that is not there
  optimizations.android.casToolchain in <home>/config.json names
  "<home>/toolchain.json ", which does not exist.
  ... and reports nothing once a file of that exact name exists.

$ stim android                            # manifest with no resourceDir
  cache       Warning: optimizations.android.casToolchain in <home>/config.json
              could not be used: <home>/toolchain.json declares no resourceDir.
              Android builds fall back to ccache when it is available.
  error       STIM_NO_METRO: No Metro port is reserved for this workspace.

$ stim android                            # the same manifest, compiler chmod 644
  cache       Warning: ... could not be used: <home>/toolchain.json names no
              executable clang, clangxx, lld, ar, ranlib. Android builds fall
              back to ccache when it is available.

$ stim doctor --platform android          # manifest `{}`, compilerCache: cas
note  A setting holds a value Stim cannot use
  optimizations.android.casToolchain in <home>/config.json could not be used:
  <home>/toolchain.json declares no clang, clangxx, lld, ar, ranlib, ndk,
  resourceDir. Android builds fall back to ccache when it is available.
  -> Repair <home>/toolchain.json, or remove optimizations.android.casToolchain
     from <home>/config.json.
costs time  The configured CMake cache predates the ccache launcher, so C++
  compiles still bypass it                # reported only because cas degraded

$ stim doctor --fix --platform android     # the same machine
  cache       removed android/app/.cxx/Debug/abc/arm64-v8a; next build reconfigures

$ stim doctor --platform android          # compilerCache: "CAS", casToolchain: "tc/toolchain.json"
costs time  Invalid optimization setting
  Invalid optimizations.android.compilerCache. Expected auto, ccache, cas, none.
                                          # and nothing naming <projectRoot>/tc

$ stim doctor --platform android          # casToolchain: null, compilerCache: auto|ccache|cas
note  A setting holds a value Stim cannot use
  optimizations.android.casToolchain in <home>/config.json is null, which is not
  an absolute path to a toolchain JSON manifest. Android builds fall back to
  ccache when it is available.
  -> Set optimizations.android.casToolchain to an absolute path to the toolchain
     JSON manifest, or remove it.
```

With `compilerCache: "none"` that last line instead reads "Android builds use no
compiler cache, because optimizations.android.compilerCache is \"none\"." An
inert key reports `A key in the config is inert` with the line that removes it.
With `config.json` truncated mid-object, both `stim doctor` and
`stim doctor --fix` print the corrupt-file note and the `mv` repair line and exit
0 instead of throwing.

Re-measured after the rebase, with the built CLI against a scratch
`STIM_HOME` and a scratch app:

- Each finding fires: a dead `casToolchain` path (`A setting points at a path
  that is not there`), `compilerCache: "cas"` with no toolchain (`A setting
  needs a companion this config does not supply`), an unread key (`A key in the
  config is inert`), and a `config.json` truncated mid-object (`The Stim config
  is not valid JSON`, exit 0).
- The full 12-case matrix of {absent manifest, manifest declaring nothing,
  `null`} x `compilerCache` {`auto`, `ccache`, `cas`, `none`} through
  `stim android`: every case warns or stays silent as designed and none refuses
  the config -- the only error reached is `STIM_NO_METRO`, from the scratch
  workspace having no reserved port.
- `doctor` and `android` print the same sentence for the same config in every
  case where both speak, including the `none` wording; for an absent manifest
  `doctor` reports the dead path instead of relaying `ENOENT`, which is the
  one-finding-per-key choice above.
- An `android.keystore` of `" release.keystore "` beside an existing
  `release.keystore` now reports nothing, and reports the trimmed path once the
  file is gone.

Guide topics and the website CAS pages were updated to match, including the
fields the manifest must name, the environment as a doctor source, and the new
repair eligibility; the guide contract tests cover the setting keys and the
`STIM_ANDROID_CAS_TOOLCHAIN` mention.

Closes #658
Closes #660


janicduplessis:
Correcting the account of finding 1, which I verified by running three builds rather than reading the code. The fix is right; the stated cause was not.

The claim was that a non-string `casToolchain` broke `android`, `ios`, `start`, `worktree warm` and `gc` alike, as a pre-existing bug. Measured, with `{ "compilerCache": "none", "casToolchain": null }` in `~/.stim/config.json`:

| build | `android` | `ios` | `gc` |
|---|---|---|---|
| published `1.0.0` | ok | ok | **refuses** |
| current `origin/main` | ok | ok | **refuses** |
| this branch, pre-fix (`0907e991c`) | **refuses**, all four `compilerCache` states | -- | refuses |
| this branch, post-fix (`89d156e63`) | ok | ok | ok |

Only `gc` was ever broken. **This branch introduced the `android` refusal**, and then fixed it. The net result is still good -- every command now degrades, and the genuine pre-existing `gc` bug is fixed too -- but the causality in the description is backwards.

The mechanism: `settingShapeErrors`' call sites are byte-identical between `main` and this branch (`caches`, `android`, `gc`, `ios`, `start`, `worktree`, `doctor`), so the gate was not widened. What moved is the declaration. `main` declares `'optimizations.android.casToolchain': 'path'` only in `optimizations.ts:17` and mentions it nowhere in `settings.ts`; this branch brought it into the shape table that `settingShapeErrors` reads. Adding the doctor sweep therefore had the side effect of making four more commands refuse a non-string value -- which is also why `SETTINGS_WITH_RESOLVE_TIME_FALLBACK` is the right shape for the fix: the key needs to be visible to the sweep and exempt from the gate.

The body should say that this branch both caused and fixed the `android` breadth, and claim only the `gc` case as pre-existing. I will rewrite it once the independent value-space audit reports, so it gets corrected once against ground truth rather than twice.

janicduplessis:
Fresh review at f8cdf7011cdfccf71a3ed5b42b8dd5f42dc03304 found one P2 blocker:

`doctor-config.ts` checks `android.keystore` verbatim, while `resolveKeystore()` trims it. With an existing `release.keystore` and `"keystore": " release.keystore "`, the build finds the file but doctor reports a missing path containing spaces. Normalize this setting consistently with its consumer while preserving the verbatim CAS path behavior.

Reproduced against the exact head. CLI build and 61 focused configuration, fallback, and doctor repair tests passed. CAS fallback and the inspected `doctor --fix` safety guards otherwise look correct. Keeping this draft pending the diagnostic fix and fresh review.

Issue #667, opened 2026-09-10
The warm lock should use the process-identity record stim already has


janicduplessis:
Blocks appandflow/stim#651 / PR #657, which is held on this.

## What happened

`worktree warm --refresh` needs a cross-process lock so a copy never reads a `node_modules` another workspace's install is rewriting. Two rounds of independent review found 17 reproduced defects, four P1 in each round, and the second round was against a lock that had already been rewritten. The findings did not converge because they were all the same shape:

- **A record and a process are two facts read at two times.** A waiter reads a claim showing no installer; the holder then records one and dies; the waiter acts on its stale snapshot and frees a claim protecting a live installer.
- **PID existence is not process identity.** A recycled PID keeps a dead claim alive until some unrelated process exits.
- **One PID is not a process tree.** Killing a package manager leaves its postinstall child writing `node_modules` while the claim is released underneath it.

## The fix is already in this repository

`packages/stim-cli/src/process-identity.ts` wraps `unique-pid` (our own package, native with prebuilds, and explicitly `No ps required`) and is already how `config.ts`, `reclaim.ts`, `metro.ts` and the collector ownership records decide whether an owner is alive. The warm lock is the one ownership record that rolled its own answer, and that is the whole bug.

Store the holder as a `ProcessRecord` (`{ pid, processToken }`) like every other owned resource, and decide liveness with `inspectProcessIdentity`, which returns exactly the four states this needs:

| status | meaning | what the lock does |
|---|---|---|
| `same` | the holder is running | wait |
| `gone` | the process is gone | reap the claim |
| `different` | the PID was recycled, so the holder is gone | reap the claim |
| `unknown` | no token, a token that does not decode, or a check that failed | refuse, and name the file to remove |

Record the installer the same way: capture its `ProcessRecord` and treat the claim as held while that record reads `same`, so a surviving postinstall child keeps the lock. Spawn it in its own process group so the group, not a single child, is the unit.

## What this removes by construction

- **No heartbeat, so no mtime, no clock, and no staleness threshold.** A holder busy in a long synchronous `simctl`, `adb` or copy call is `same`, not stale, so it can never be falsely reaped. This was the likeliest failure of the heartbeat design, and stim's own guide already warns that such calls hold its timers.
- **No `ps`.** It was unavailable under a sandbox before, which is why `unique-pid` exists.
- **PID reuse is `different`, not a live-looking claim** -- the second round's finding 8, gone outright.
[...]

PR #681, opened 2026-09-11
feat: one ownership-claim primitive for long-held locks

### LINKED ISSUE #678: One ownership-claim primitive for long-held locks, backed by process identity
## Two kinds of lock, one shared implementation

`withDirLock` in `@stim-cli/core` guards **short critical sections**: read a JSON file, modify it, write it back. Milliseconds. Six modules share it (`config.ts`, `paths.ts`, `supervisor/state.ts`, `diagnostic-store.ts`, `eas-session-ledger.ts`, `device-lease.ts`), and its mtime staleness is appropriate at that duration.

The other kind is an **ownership claim held for the duration of a long operation** -- a build, an install, a supervisor's lifetime. Those were written from scratch three times, and a fourth is in review:

| module | lines | record | liveness |
|---|---|---|---|
| `engine/build-lock.ts` | 342 | `{ pid }` | `isPidAlive` + mtime + rename-aside reap |
| `engine/build-slots.ts` | 240 | `{ pid }` | `isPidAlive` + mtime |
| `engine/workspace-process-lock.ts` | 134 | `{ pid }` | `isPidAlive` + mtime |
| `engine/warm-lock.ts` (PR #657) | -- | `{ pid, installerPid }` | `isPidAlive` |

Long duration is exactly when a bare PID check and mtime staleness go wrong, and two independent reviews of the warm lock found **17 reproduced defects** in this pattern -- four P1 in each round. The same shapes are present in the three that already shipped.

## Meanwhile the safe primitive is right there

`packages/stim-cli/src/process-identity.ts` wraps `unique-pid` (ours, native, `No ps required`) and returns exactly the four states an ownership claim needs:

| status | meaning | claim behaviour |
|---|---|---|
| `same` | the holder is running | wait |
| `gone` | the process is gone | reap |
| `different` | the PID was recycled, so the holder is gone | reap |
| `unknown` | no token, or it will not decode | refuse, naming the file to remove |

`config.ts` and `supervisor/state.ts` already store `{ pid, processToken }` and decide this way. The long-held locks are the ones that did not.

## What to build

One ownership-claim primitive, with the claim semantics that the reviews showed are required:

- **Build the claim privately and rename it into place**, so a half-built claim is never observable. No time-based takeover.
- **Address the claim by a token in its own filename**, so a reader that observed a dead claim can only ever unlink that exact claim. Re-read immediately before unlinking.
- **Store a `ProcessRecord`** and resolve liveness through `inspectProcessIdentity`. No `isPidAlive`, no mtime, no staleness threshold, no heartbeat -- so a holder busy in a long synchronous `simctl`, `adb`, or copy call reads `same` rather than stale and can never be falsely reaped.
- **Shared and exclusive modes**, since the warm lock needs many copies against one refresh and `build-slots` needs counting.
- **An optional child record**, for a claim whose real writer is a spawned process (a package manager that outlives its parent). Spawn in its own process group and treat the group as the unit.
- **Refuse, never silently wait, when the state cannot be established**: `unknown`, a child marker with no usable record, or a reap that fails with anything but `ENOENT`. Print the claim path and the command to remove it, the way git does for `index.lock`. Swallowing a failed reap is what produced the ninety-minute waits.

## Then migrate

`build-lock`, `build-slots` and `workspace-process-lock` onto it, each keeping its current observable behaviour except where that behaviour was the bug. Every one of them is load-bearing, so migrate one per pull request with its existing tests kept green, rather than in one sweep. `build-lock` first: it is the single-flight that stops several agents compiling the same app.

PR #657's warm lock then becomes the first new consumer rather than a fourth implementation; see #667.

## And make the wrong call hard to reach

`isPidAlive` is a bare `process.kill(pid, 0)` used in about 25 modules. It is **correct** for "is the child I spawned still running" (`start.ts`, `launch.ts` hold the handle) and **wrong** for "is the process that wrote this record on disk still that process". One name, two meanings, and the locks picked the wrong one four times. Separate them so the record-based question cannot resolve to the PID-existence answer -- at minimum a name that says what it does not guarantee.

Also audit, rather than assume, the record-based callers that are not locks: `commands/gc.ts` decides what to **delete** from a bare PID check (two sites), and `engine/device.ts` decides device ownership the same way (four sites).

## A decision to record before implementing

`process-identity.ts` lives in `stim-cli`; `withDirLock` lives in `core`, which is dependency-light. `unique-pid` is a native module with prebuilds. Decide deliberately whether the new primitive lives in `stim-cli` beside `process-identity`, or in `core` with the native dependency moving too -- and say why in the pull request.

## Which of the two to use

Keeping both is deliberate, because a wrong answer costs the opposite thing in each case and so they want opposite recovery behaviour:

- A `withDirLock` critical section is a few milliseconds. A dead holder has to clear in seconds or everything jams, and a wrong reap is nearly harmless -- the write it guards is atomic anyway, so you redo it. Timeout-based recovery is right.
- An ownership claim is held for a build. A wrong reap means two compiles racing into one `node_modules`, and waiting is cheap next to that. It must never reap unless certain, which is what process identity and a refusal posture buy.

They also differ in three ways that resist merging. `withDirLock` is reentrant on purpose, because config read-modify-write nests and re-entry must be a no-op; an ownership claim taken twice is a bug, not a no-op. Process identity is free once per build and wasteful on a guard that runs hundreds of times per command. And `remove this file manually` is the right message for a stuck build lock and a terrible one for a config write.

**The rule, which belongs in a comment on each primitive so the next author picks correctly:** if you would print the holder to a human -- `/w/app-412 is already building a3f9b1.. (pid 41233)` -- it is an ownership claim. If you would just wait your turn, it is a dir lock.

## Out of scope

`withDirLock` itself, beyond adding that selection note. Short critical sections are not the problem, and changing them would touch six modules for no defect.

janicduplessis:
## Description

Three modules held an ownership claim for the length of a long operation -- a compile, a slot in the
build semaphore, a Metro start -- and each one implemented that claim from scratch: `engine/build-lock.ts`
(342 lines), `engine/build-slots.ts` (240), `engine/workspace-process-lock.ts` (134). All three stored a
bare `{ pid }` and decided liveness with `isPidAlive` plus an mtime staleness threshold. A fourth
implementation of the same pattern (the warm lock, #657) drew two independent reviews that reproduced 17
defects, four P1 in each round, and they did not converge because every one was a property of the pattern:

- a bare pid cannot tell a recycled pid from the process that wrote the record, so a dead claim stays
  unreapable until an unrelated process exits;
- mtime staleness falsely reaps a holder busy in a long synchronous `simctl`, `adb` or copy call --
  exactly the calls Stim's own guide warns hold its timers;
- a reap on a fixed path can remove a claim that replaced the one the reader observed;
- a swallowed failed reap turns into a 90-minute silent wait.

`packages/stim-cli/src/process-identity.ts` already answers the question these locks were asking. It wraps
`unique-pid` (ours, native, no `ps`) and returns `same` | `gone` | `different` | `unknown` for a
`ProcessRecord`; `config.ts` and `supervisor/state.ts` already decide ownership this way. The long-held
locks were the ones that rolled their own answer.

Blast radius: every `stim ios` / `stim android` build (the single-flight lock and the `maxBuilds`
semaphore), `stim start` (the `metro-start` lock), the managed tunnel, remote-session and EAS project
locks, `gc`, and `doctor`. `withDirLock` in `@stim-cli/core` is untouched: its mtime staleness is correct
for a millisecond critical section, and no defect points at it.

## Solution

One primitive, `packages/stim-cli/src/ownership-claim.ts`, then three migrations, one commit each.

**Where it lives, and why.** In `stim-cli`, beside `process-identity.ts`. The alternative was `core`, next
to `withDirLock`, but `core` declares no runtime dependencies at all and is the package the cache plugins
(`@stim-cli/cache`, `metro`, `expo-build-cache`) load when Stim is not installed. Moving the primitive
there means moving `unique-pid` there, pushing a native module with prebuilds onto every consumer of a
dependency-light package for a primitive none of them use. Every consumer of the claim is in
`stim-cli/src/engine/`, and the liveness answer it is built on is already in `stim-cli`.

**Shape of a claim.** A claim set is a directory. An exclusive claim is published by building
`<root>/.staging-<uuid>/<claimId>.claim` privately and renaming that *directory* onto `<root>/exclusive`:
rename onto a non-empty directory fails `EEXIST`/`ENOTEMPTY`, so the rename is the mutual exclusion, and
the slot is complete from the instant it is observable -- there is no window in which a claim is half
built, and therefore no age threshold to cover one. Shared claims are files under `<root>/shared/`, each
named by its own token.

Exclusion between the modes is publish-then-verify with a deterministic tie-break: a reader that sees a
live exclusive claim withdraws its own and waits; a writer that publishes its slot and then finds live
readers *keeps* the slot and reports `waitingFor`, which blocks new readers and drains the existing ones.
Writer preference, so neither side can livelock and a reader stream cannot starve a writer.

A claim set is removed the moment it holds nothing, so every step of a publication can lose the directory
it is writing into to another process's cleanup. `EEXIST`, `ENOTEMPTY`, `ENOENT` and `EINVAL` anywhere in a
publication mean "try again" -- never "nobody holds this" -- and the acquire loop recreates the set and
retries, bounded. This matters more than it looks: without it the lock raises a raw errno that
`ios`/`android` catch as "building anyway", silently bypassing the single flight under exactly the
contention it exists for.

**Liveness.** A claim stores the holder's `ProcessRecord` and is resolved only through
`inspectProcessIdentity`: `same` is held, `gone` and `different` are reaped, `unknown` is refused. No
heartbeat, no mtime, no staleness threshold, no clock anywhere in the decision. A claim may also carry a
child record for a claim whose real writer is a spawned process that can outlive its parent: marked
pending *before* the spawn, re-read after the owner is known dead, and resolved as held while the recorded
leader reads `same`, or while `gone` and its process group still has a member -- so a package manager's
postinstall child keeps the claim after the manager exits. A child whose identity reads `different` is a
refusal, not a group check: the pid was recycled, so `-pid` no longer names our group, and guessing either
way would either free a live claim or keep a dead one alive until an unrelated process exits.
`detached: true` is part of that contract and is documented on `setClaimChild`.

**Reaping.** Because a claim is addressed by a token in its own filename, a reader that decided claim A is
dead can only ever unlink claim A: it re-reads the file immediately before unlinking and aborts when the
id changed, and the `rmdir` that follows fails harmlessly on a directory a new claim has already entered.
The child marker is removed before the claim record, so an interrupted reap leaves a complete,
still-resolvable claim rather than an orphan nothing can resolve.

**Refusals, and what an unrecordable identity means.** Four states cannot be established, and each is an immediate refusal that changes nothing:
an `unknown` identity, a child declared but never recorded, a claim path that is a regular file, and a
reap that fails with anything but `ENOENT`. The error names the refused claim and the exact removal for
*that* path -- git's posture on `index.lock`, which is why git has none of these races. The removal is
shell-quoted through `quotedPath` and targets the claim's own file rather than the directory around it: this
is text printed for a human to copy into a shell, so an unquoted path with a space in it would name several
targets, and a single claim record does not need the lock directory removed. One command is
enough to clear one claim: a child marker left without its claim record holds nothing, so the next pass
removes it rather than refusing a second time over it. `ios` and `android` turn it into
`STIM_CLAIM_REFUSED` with that command as the remedy instead of their previous "building anyway", which
would have swallowed it; `guide errors STIM_CLAIM_REFUSED` documents it.

`STIM_CLAIM_UNAVAILABLE` is a refusal at every call site, deliberately and uniformly: `selfOwner` cannot
capture a process identity (no `unique-pid` prebuild for this platform, or the OS refusing to report this
process's start identity), so no claim can be written at all. The alternative -- proceed loudly when no
other claim exists -- is not safe: the condition is a TOCTOU, because a competing run finds the same empty
set and also proceeds, and a warning does not prevent the duplicated compile it warns about. A run with no
claim is invisible to every other run, so the single flight and `concurrency.maxBuilds` would both be off at
once. It also matches what `collector/run.ts` and `supervisor/run.ts` already do with the same failure
("refusing to start an unmanaged collector"). `claimFailure` maps both claim codes to a refusal at all six
`ios`/`android` sites, so an unrecognised claim error can no longer reach the "building anyway" path; the
refusal names the `unique-pid` error code and message. The residual `catch` there still warns and builds for
a raw filesystem errno on `$STIM_HOME` (`EACCES`, `EROFS`, `ENOSPC`), which is a broken lock store rather
than an undecided claim, and refusing every build on a read-only `$STIM_HOME` would be worse.
`tryAcquireBuildSlot` keeps catching only `STIM_CLAIM_REFUSED` to skip one slot, since an unavailable
identity applies to every slot and must not be retried per slot. `guide errors STIM_CLAIM_UNAVAILABLE`
documents it, and a contract test holds both claim codes to the errors topic.

**Clearing a claim set.** `gc --delete` no longer removes a claim set with a recursive delete. It calls
`clearFreeClaimSet`, which takes the set's own exclusive claim first and then removes the set only through
the token-verified unlink and empty-directory removals that releasing a claim performs. A claim taken
between the survey that reported the set stale and the delete therefore survives: the `rmdir`s fail
harmlessly on a directory that claim has entered, and gc reports the set as held and leaves it. A set that
has become unresolvable by the time gc reaches it is refused and named, not deleted -- treating `unresolved`
as "do not delete" at the point of deletion and not only at the point of reporting. An interrupted
publication's `.staging-` leftover is removed while the claim is held, so a set holding only crash debris
still clears; anything else left in the set is a delete failure naming what remains.

Nothing reaps an unresolved claim, `gc --delete` included: it is excluded from `stale`, reported under
"Build locks Stim cannot resolve - NOT touched", and left for the user. Invariant 8 and the fact that one
transient failure inside `unique-pid` would make *every* claim read unresolved at once make the automated
sweep the wrong place for that judgement. `doctor-cxx` treats an unresolved Android lock as an active
build for the same reason. A build slot whose holder cannot be identified is skipped while any other slot
is merely busy, and refuses only when no slot is left to wait for.

**Mixed Stim versions, recorded deliberately.** The claim paths are unchanged, but the record inside them
is not, and neither direction of a version skew is safe: a new Stim reads an older `lock.json` claim set as
free, and an older Stim reads a new claim set as a lock with no record and *deletes* it after its 5s grace.
There is no compatibility shim here on purpose -- writing a decoy `lock.json` would reintroduce the
duplicated record format this change exists to delete, and a path rename would only convert a destroyed
claim into two concurrent builders. What a skew costs is one duplicated operation per lock rather than
corrupted state, but "one operation" now spans everything these locks guard: a duplicated compile, a
duplicated Metro start, a duplicated managed tunnel or worktree-removal race, and a duplicated
remote-session claim, which on `--remote eas` means a second billed cloud simulator session. The release
note for this change should say not to run two Stim versions concurrently against one machine -- not just
one `$STIM_HOME`, since the `eas-project` lock lives under `homedir()/.stim/machine/eas`, which
`STIM_HOME` does not move.

**Audit of the record-based callers that are not locks.**

- `commands/gc.ts` (two sites) was genuinely wrong: it re-read a build lock and a build slot record and
  decided to **delete** from `isPidAlive(record.pid)`, so a recycled pid protected a dead lock and -- worse
  -- the guard passed for a record it could not parse. Both now re-survey the claim set, skip any lock with
  a live claim, and never see an unresolved one. Fixed here because the migration already replaced the
  records they read.
- `engine/device.ts` (four sites) is **not** a record-based caller. All four `alive = isPidAlive` defaults
  flow into `emulatorGone(pid, alive)`, where `pid` is the pid `bootAndroidEmulator` returned in this run,
  used to abort a boot wait when the emulator process exits. That is the question a bare pid check answers
  correctly. Device *ownership* is decided by the `stim-` name and the `owned: true` record, never by a pid.
  No change.
- `reclaim.ts` and `collector/ownership.ts` read records, but the pid check is only a prefilter in front of
  `verifyCollectorOwnership`, which decides with `inspectProcessIdentity`; PID existence there can only
  produce `unverified` ("keep the record, signal nothing"), never ownership. No change.

**Making the wrong call harder to reach.** `isPidAlive` is now `pidExists` across all 17 files that used
it, with a doc comment naming what it cannot answer and where to go instead. The three migrated locks no
longer take a liveness seam at all: there is no `isAlive` option to pass the wrong function to, and the
record-based question resolves only through `inspectProcessIdentity`.

## Test plan

`packages/stim-cli/src/__tests__/ownership-claim.test.ts` (25 tests) drives each state with a real or
rebuilt `ProcessRecord` rather than a stub: a live holder, a holder killed with `SIGKILL` in a separate
process, a pid recycled under a different identity, a token that does not decode, a truncated record, a
child marker with no record, a recycled child pid, a reap that cannot remove the file, a claim root that
is a file, a remedy that has to clear a claim with a child marker in one step, six real processes
racing one claim over a crash-left claim, four processes cycling the same claim set 150 times each in both
modes with an overlap witness inside the critical section, and -- with real processes -- a detached leader
whose own child outlives it, where the claim stays held until the process group is empty and is reaped
immediately after. Age independence is asserted directly: a claim back-dated six hours whose holder runs
is still held.

Contention evidence, run outside the suite against the real modules (these are the reviewer's
reproductions, re-run after the fix):

| probe | before | after |
|---|---|---|
| 6 processes x 300 exclusive cycles, 2 rounds | 148 raw errnos escaping `tryAcquireClaim` | 12/12 processes clean, 643 acquisitions, 0 overlaps |
| 6 processes x 400 shared cycles | 5 of 6 processes died by iteration 12 | 6/6 x 400 acquisitions, 0 errors |
| 8 processes x 400 `withWorkspaceProcessLock` cycles | 12 raw `ENOENT` | 8/8 x 400 cycles, 0 errors |

Migrations, in order, each with the module's existing tests kept green:

- `engine/build-lock.ts` -- 36 tests pass, including all three real-process tests unchanged (six-process
  race, builder/waiter handshake, `SIGKILL`ed builder taken over).
- `engine/build-slots.ts` -- 13 tests pass, including the live 2-slots/3-processes handoff unchanged.
- `engine/workspace-process-lock.ts` -- covered through its callers (`start`, tunnel, remote-session, EAS
  project, worktree removal); no test needed an assertion change, only the two fixtures below.

Tests whose assertions or fixtures changed, and why:

- `engine-build-lock.test.ts` "a lock held by a DEAD pid is taken over" asserted that `{ pid: 999999 }`
  with an injected `isAlive: () => false` is taken over. That assertion encodes the defect: the record
  carries no process identity, so the new lock cannot prove the holder is gone and pid 999999 may belong
  to someone else. It is now "a lock whose builder is GONE is taken over", driven by a real gone record,
  plus two new cases -- a recycled pid is taken over, and an unidentifiable holder refuses while leaving
  the claim in place.
- The same file's "a lock with no lock.json at all is taken over once it has aged" asserted the 5s grace
  window. A lock directory with no claim in it is now simply free, so the test asserts that for both a
  fresh and a six-hour-old directory -- the fresh case is the one the old grace blocked.
- Fixture-only rewrites, assertions preserved: the `lock.json` / `slot.json` / `owner.json` writers in
  `engine-build-lock.test.ts`, `engine-build-slots.test.ts`, `gc.test.ts` and `worktree-remove.test.ts`
  now plant claims through shared `_factories` helpers (`liveClaimOwner`, `goneClaimOwner`,
  `recycledClaimOwner`, `plantClaim`). `listBuildLocks` / `listBuildSlots` lost their `isAlive` option, so
  the two tests that injected it now plant live and gone claims instead.
- `gc.test.ts` "a malformed remote session lock disables the EAS sweep" planted a malformed `owner.json`.
  `commands/gc/eas-sessions.ts` was parsing that record itself -- a fourth copy of the format -- and now
  asks the claim survey; the fixture plants a malformed claim. Behaviour changes with it: a
  `remote-session.lock` directory that exists but holds no claim used to produce the malformed notice and
  now reads as free, which is what it is.
- `gc.test.ts` / `gc-output.test.ts`: the report's `buildLocks` and `buildSlots` gain an additive
  `unresolved` array, so the exact-shape assertion and the serialized-shape snapshot were updated. No
  documented payload key changed meaning.

Three regression tests for the second review round, each driven the way the reviewer drove the defect:

- `gc.test.ts` spawns a real `runGc --delete` paused inside the first removal it makes under the claim set
  it decided was stale, for a build lock and for a build slot. A second real process takes the set in that
  window; after gc finishes, that process's claim file still exists, it is the only live claim, a third
  acquirer is held off by it, and gc printed no "Cleared". A third test plants a claim that became
  unresolvable in the same window and asserts it is still there and gc said it left it alone. All three fail
  against the previous recursive delete -- the live and the unresolvable claim are both gone.
- `ios-command.test.ts` / `android-command.test.ts` inject `ClaimUnavailableError` through each harness at
  the build-lock site and the build-slot site: `ok: false`, the `STIM_CLAIM_UNAVAILABLE` code, a remedy
  naming `unique-pid`, no build, and no "building anyway". All four report `ok: true` and one completed
  build if the code is not mapped to a refusal. A fifth drives `ClaimRefusedError` through the same mapping.
- `ownership-claim.test.ts` runs the printed remedy as printed, through `getExecutor().run`, against a claim
  in `<home>/cache home/build.lock` and `<home>/it's cache/build.lock` with a `<home>/cache` sibling holding
  a file. The claim goes, the sibling stays. Unquoted, the space case deletes the sibling and leaves the
  claim. Two more cover `clearFreeClaimSet` directly: a set left by an interrupted publication clears while
  content Stim did not write is kept and reported, and a set a live process holds is untouched.

Gates from the repo root, outside the sandbox, on `cbdceb33b` (rebased onto `da6fe8431`):
`pnpm format:check`, `pnpm lint`, `pnpm typecheck`, `pnpm knip`, `pnpm test` (3933 passed, 1 skipped, 125
files), `pnpm test:e2e` (78 passed, run repeatedly). The claim suite was also run repeatedly on its own to shake the races out. No
real-tool invocation changed, so invariant 9 does not apply; the process-level behaviour is exercised
against real spawned and killed processes rather than mocks.

Closes #678. #667 (the warm lock) becomes the first consumer of this primitive rather than a fourth
implementation, and lands on top of it.


janicduplessis:
Fresh review at cbdceb33b9766d54eab742b8612ee2ceac57d8fd found two P2 blockers:

1. `clearFreeClaimSet()` calls `removeAbandonedStaging()` without guaranteed claim release. A populated `.staging-abandoned` directory with mode `000` causes raw `EACCES` and leaves the newly acquired claim live. The GC callers no longer catch per-entry filesystem errors, so this aborts the remaining lock, slot, lease, and cache cleanup. Reproduced using real filesystem permissions. Ensure claim release on failure and contain the failed entry so the sweep continues.
2. The `metro-start` process lock now propagates claim refusals, but `startLocked()` only maps timeouts. Running `start --json` with a malformed claim produced empty stdout and an uncaught `ClaimRefusedError` stack. Route claim failures through the existing structured failure handling so stdout contains one JSON payload.

The focused ownership-claim suite passed 28/28. Keeping this draft pending fixes and fresh review.

PR #688, opened 2026-09-11
fix: make late remote-start refusal test deterministic

### LINKED ISSUE #643: Make the late local-supervisor remote-start test deterministic
## Problem
The Node 22 CI test "a remote start refuses after the local Expo supervisor begins answering without a tunnel" returned STIM_METRO_TIMEOUT instead of STIM_REMOTE_START_REQUIRED. It passed on Node24 and locally at the same commit.

## Evidence
CI run https://github.com/appandflow/stim/actions/runs/34505875838 , Node22 job102967819147, start.test.ts:2162, head b4a313c912daed63199e930583fe48849daa7eb9. PR642 changes only benchmark audit files/docs; the affected test and start implementation are unchanged from current main26d749d1a508b1f7c748430d70f32e82f9b716bd. Only the failed job was retried, without changing its assertion.

## Cause
Not fully diagnosed. The test starts a real listener with metroListener(port).then(...) without awaiting it and invokes runAction with wait=1 second. Scheduling or health-poll timing under CI load may exhaust this deadline before the intended remote-mode refusal.

## Fix idea
Synchronize the delayed-listener transition explicitly or use a deterministic clock/readiness boundary. Preserve coverage that start waits for the existing supervisor, never spawns another, then refuses its incompatible remote mode. Do not simply accept either timeout or refusal.

janicduplessis:
## Description

The late local-supervisor test reported `STIM_METRO_TIMEOUT` on Node 22 CI instead of `STIM_REMOTE_START_REQUIRED`. Its unawaited listener competed with a one-second deadline, so scheduling could prevent it from reaching the remote-mode refusal.

## Solution

Control the test clock and await the listener after Stim enters the existing-supervisor wait. Advance the tunnel deadline only after the real HTTP health probe succeeds. The test also asserts that it took the waiting path and still requires the exact refusal code and no supervisor spawn. Production code is unchanged.

## Test plan

- 30 repeated focused runs on Node 22.22.2 passed.
- Full `pnpm test` passed.
- `pnpm run format:check`, `pnpm run lint`, `pnpm run build`, `pnpm run typecheck`, and `pnpm run knip` passed.

Fixes #643

### DIFF
--- a/packages/stim-cli/src/__tests__/start.test.ts
+++ b/packages/stim-cli/src/__tests__/start.test.ts
@@ -2127,10 +2127,14 @@ describe('action: an existing supervisor that is not answering', () => {
     );
     const port = 8167;
     const exec = metroExecutor({ listeners: {} });
-    const held: { server: Server | null } = { server: null };
+    let markMetroHealthy: () => void;
+    const metroHealthy = new Promise<void>((resolve) => {
+      markMetroHealthy = resolve;
+    });
     const base = exec.runQuiet.bind(exec);
     exec.runQuiet = (cmd) => {
       if (new RegExp(`lsof -nP -iTCP:${port}`).test(cmd)) return exec.listening ? '5153' : '';
+      if (/lsof -a -p 5153 -d cwd/.test(cmd)) markMetroHealthy();
       return base(cmd);
     };
     setExecutor(exec);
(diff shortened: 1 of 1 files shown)


janicduplessis:
Two fresh independent in-session reviews are clear at f5eb4a2ae59e52d936179556da7c53629a1920e9. The second review identified a Linux path where a real mocked PID could bypass the lsof readiness signal; both reviewers verified that the final impossible-PID fixture resolves it. The test preserves the delayed-supervisor wait, real HTTP health check, exact remote-mode refusal, restart remedy, and no-spawn assertion. Production code is unchanged.

All local checks passed, including 3,891 unit tests (1 skipped) and 30 repeated focused Node 22 runs after the correction. CI passed for Node 22, Node 24, and published Node 20.19.4 on this exact head. No actionable findings remain.

External Claude review was blocked by automatic approval review because it would send repository content to another service; both independent reviews used the explicitly authorized in-session agents.

PR #692, opened 2026-09-11
fix: distinguish owned devices across worktrees

### LINKED ISSUE #684: Owned iOS simulator names collide across worktrees; Android already disambiguates
## Problem

Two workspaces for the same app get the same owned iOS simulator name. Android does not have this problem, because its creation path already disambiguates.

Stim itself is unaffected — it resolves its own device by UDID from the project record — so the cost lands on everything downstream that selects a device by name, and on a human reading `xcrun simctl list`.

## Evidence

Two worktrees of the same app, both registered:

```
tlon-mobile [idle] (/Users/.../tlon-apps/.worktrees/login-button-test/apps/tlon-mobile)
  ios: stim-tlon-mobile (iPhone 17 26.5) shutdown (owned)
  android: stim-tlon-mobile (emulator) (owned)

* tlon-mobile (/Users/.../tlon-apps/.claude/worktrees/pr6460/apps/tlon-mobile)
  ios: stim-tlon-mobile (iPhone 17 26.5) booted (owned)
  android: stim-tlon-mobile-d6ba72aa (emulator) (owned)
```

The second workspace's Android AVD picked up a suffix. Its iOS simulator did not, and collides exactly with the first.

`agent-device devices --platform ios` then lists the name twice, and selecting by it fails:

```
stim-tlon-mobile (iPhone 17 26.5) (ios simulator target=mobile) booted=true
stim-tlon-mobile (iPhone 17 26.5) (ios simulator target=mobile) booted=false

Error (DEVICE_NOT_FOUND): No device named stim-tlon-mobile
```

## Cause

`packages/stim-cli/src/engine/device.ts:648-652` disambiguates only on the Android branch:

```ts
if (
  readParked('android').some((entry) => entry.name === ownedAvdName(label)) ||
  findOtherProjectOwningAvd(ownedAvdName(label), projectPath)
)
  label = `${label}-${randomUUID().slice(0, 8)}`;
```

The iOS path builds its name straight from the label with no collision check — `ownedSimName(label, model)` at `device.ts:339` and `device.ts:374`. `randomUUID` is imported once in that file and used only in the Android branch above, and `findOtherProjectOwningAvd` (`device.ts:475`) has no iOS counterpart.

The label is the workspace label, which is the app directory name, so every worktree of one app wants the identical name. `ownedSimName` does carry a `suffix` parameter, but only `parkedSimName` passes one.

## Fix idea

Give the iOS creation path the same collision check Android already has, so the two platforms behave alike and a name is a usable selector on both.

`fitSimName` already accepts a `suffix` and already accounts for its length against `SIM_NAME_MAX`, so the shape exists; what is missing is an iOS equivalent of `findOtherProjectOwningAvd` plus the parked-name check, and passing the suffix through `ownedSimName`.

Alternative considered and not preferred: document that names are not unique and require the reported device ID everywhere. `guide agent` already says to use the full reported device ID and never assume a simulator named booted belongs to this workspace, and that guidance did not prevent the collision above from being hit. Making the name reliable seems better than repeating that it is not, especially since Android already sets that precedent.

Closing this needs evidence that two worktrees of one app produce two distinct iOS simulator names, and that an existing single-workspace name is unchanged.

janicduplessis:
## Description

Worktrees of the same app currently get identical iOS simulator names, making name-based device selection ambiguous. Local device labels now combine the worktree and app directory names:

```text
pr6460/apps/tlon-mobile            -> stim-pr6460-tlon-mobile (iPhone 17 26.5)
login-button-test/apps/tlon-mobile -> stim-login-button-test-tlon-mobile (iPhone 17 26.5)
myapp/                            -> stim-myapp (iPhone 17 26.5)
```

## Solution

Basenames keep their existing casing and punctuation (`My.App` stays `My.App`). If the formatted iOS name still collides, append the workspace ID after the model and runtime so truncation cannot remove it. Creation, reuse, and parked adoption share the existing process-held lock to prevent concurrent name assignments. Booting remains outside that lock. New Android AVDs use the same readable labels; existing and adopted AVD names stay unchanged.

## Test plan

- Regression coverage uses real git worktrees and a symlink; checks separate apps, equal basenames, and directories outside git. Device tests cover concurrent allocations, truncation, unavailable and parked names, adoption, and reuse.
- Real `simctl` on iOS 26.5 verified the three names above and a 60-character rename retaining the workspace ID. Calls had 30-second timeouts; all three disposable simulators were deleted through Stim teardown, without booting or building an app.

Fixes #684.

### DIFF
--- a/packages/stim-cli/src/__tests__/engine-device.test.ts
+++ b/packages/stim-cli/src/__tests__/engine-device.test.ts
@@ -17,6 +17,8 @@ import { allConsolePortsAndSerials, getProject, setDevice, upsertProject } from 
 import type { DeviceRecord } from '../types.ts';
 import { resetExecutor, setExecutor } from '../exec.ts';
 import { parkSim, readParked } from '../sim-pool.ts';
+import { workspaceId } from '../paths.ts';
+import { ownedSimName } from '../sim/ios.ts';
 import { makeAdbDevices, makeChildProcess, makeConfig, makeExitingChild, makeIosSim } from './_factories.ts';
 
 type SimEntry = {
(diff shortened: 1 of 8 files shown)


janicduplessis:
Two independent in-session reviews are clear at 5a49843e1d7ab1544cdbcb1e7df1414df427b2e2. They covered label derivation, collision handling, creation/reuse/adoption locking, regression tests, and guidance; no actionable findings. Reviewer validation included the 102-test device/project suites.

Local validation passed: 3,897 unit tests, one skipped; 78 end-to-end tests; formatting, lint, build, typecheck, knip, and runtime checks. Real timed simctl creation/rename checks passed and the disposable simulators were removed through centralized teardown. CI is green for Node 22, Node 24, and the published Node 20.19.4 runtime.

PR #693, opened 2026-09-11
fix: tolerate unavailable iPhones in the live probe test

### LINKED ISSUE #631: Live devicectl process-probe test fails when the attached iPhone is unreachable
## Problem

Running `pnpm test` on a Mac with an iPhone attached fails one test:

```
FAIL packages/stim-cli/src/__tests__/engine-ios-device.test.ts
  > listIosDevices against a real devicectl
  > the process-probe argv is accepted by the real devicectl when a phone is connected
AssertionError: expected undefined to be null
  engine-ios-device.test.ts:700
```

The test asks for the process of an app that cannot exist and expects `null`, meaning "the probe worked and found nothing". It got `undefined`, meaning "the probe could not run". The rest of the suite passed, and the failure is unrelated to the change being tested at the time (website styling, #630).

## Evidence

Observed 2026-09-10 on macOS 25.6.0 while running the full suite from `origin/main` at 6789520 plus website-only changes. The first live test in the same `describe`, which only lists devices, passed, so `devicectl` itself runs and enumerates the phone.

## Cause

Partly known. `iosDeviceProcess` in `packages/stim-cli/src/engine/ios-device.ts:341` returns `undefined` from its catch block whenever `devicectl device info processes` throws, which includes a timeout and any refusal from a locked, untrusted, or otherwise unreachable phone. The test at line 700 treats a connected phone as a reachable one. Why the phone refused the process listing in this run was not diagnosed.

## Fix idea

Make the live test distinguish "device unreachable" from "probe broken": accept `undefined` only when `devicectl` reports a refusal the code already classifies (the `IosLaunchRefusalKind` patterns), and keep failing on `undefined` for any other reason. That preserves the argv check the test exists for and stops a locked phone on a developer's desk from failing the suite.

janicduplessis:
## Description

`devicectl` lists remembered iPhones that are currently unavailable. The live process-probe test treats those devices, or a locked phone, as a broken probe and fails `pnpm test`.

## Solution

Accept existing classified launch refusals and the exact CoreDevice message/code pairs observed here: 4000 for an unavailable connection and 1011 for a device that cannot be found. Verify that the command receives the selected UDID before accepting any refusal. Unknown errors, child-process timeouts, and missing output still fail; production code and command arguments are unchanged.

## Test plan

- `pnpm test packages/stim-cli/src/__tests__/engine-ios-device.test.ts`: 50 passed with real `devicectl` enabled and an isolated `TMPDIR`.
- Full `pnpm test`: 124 files passed, 3,892 tests passed, 1 skipped.
- A temporary executable fixture covered successful probes, recognized refusals on stdout/stderr, malformed errors, timeouts, missing output, and incorrect arguments or UDIDs.
- `pnpm run format:check`, `pnpm run lint`, `pnpm run build`, `pnpm run typecheck`, `pnpm run knip`, and `pnpm run test:runtime` passed.

Fixes #631

### DIFF
--- a/packages/stim-cli/src/__tests__/engine-ios-device.test.ts
+++ b/packages/stim-cli/src/__tests__/engine-ios-device.test.ts
@@ -698,6 +698,7 @@ describe('listIosDevices against a real devicectl', { skip: LIVE as unknown as b
     if (!connected) return;
     const executor = getExecutor();
     let failure = '';
+    let timedOut = false;
     let probedUdid: string | undefined;
     const pid = iosDeviceProcess(
       { udid: connected.udid, appName: 'NoSuchAppStimWouldEverBuild' },
@@ -709,8 +710,9 @@ describe('listIosDevices against a real devicectl', { skip: LIVE as unknown as b
             try {
               return executor.runFile(file, args, options);
             } catch (error) {
-              const failed = error as Error & { stderr?: unknown; stdout?: unknown };
+              const failed = error as Error & { stderr?: unknown; stdout?: unknown; code?: unknown };
               failure = [failed.message, failed.stderr, failed.stdout].filter(Boolean).join('\n');
+              timedOut = failed.code === 'ETIMEDOUT';
               throw error;
             }
           },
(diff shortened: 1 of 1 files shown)


janicduplessis:
Both independent in-session reviewers are clear at 764abb4c0248df0baa6bdbfe1ea3830e40affc54. The final test rejects ETIMEDOUT before classifying refusal text, including a recognized locked-device response followed by a hung child process. No remaining actionable findings.

Validation passed: 3,892 unit tests, one skipped; formatting, lint, build, typecheck, knip, and runtime checks. Executable fixtures verify successful probes, recognized refusal responses, strict CoreDevice availability pairs, unknown and malformed errors, missing output, wrong UDIDs, and timeouts with or without refusal output. The real devicectl-focused suite passed 50 tests. Node 22, Node 24, and published Node 20.19.4 runtime CI are green.

PR #694, opened 2026-09-11
fix: preserve compiler cache statistics after launch failures

### LINKED ISSUE #644: Report compiler cache statistics before a later launch failure

janicduplessis:
## Description

Android's compiler-cache counters disappear from command output when installation or launch fails. Retrying from the artifact cache then reports only that compilation did not run. Both platforms also discard available counters from JSON failures.

Fixes #644.

## Solution

Print the measurement once on stderr when the build returns, and include available counters in JSON errors from failed builds or later install/launch steps. Remove the duplicate measurement from successful stdout summaries; artifact hits retain the `not run` line.

This relaxes the [stdout-summary rule](https://github.com/appandflow/stim/commit/e759551543e7ca0b22a29f99906ac167545bcdc7) so a later failure cannot hide a completed measurement, using the existing `cache` progress label.

## Test plan

- Command regressions verify counters survive failed builds, installation and launch, appear before installation, and do not repeat on success. JSON failures remain one payload.
- Build-engine regressions preserve counters from Xcode metrics and fresh ccache logs across successful and failed exits. Guide coverage checks the JSON failure fields.

### DIFF
--- a/packages/stim-cli/src/__tests__/android-command.test.ts
+++ b/packages/stim-cli/src/__tests__/android-command.test.ts
@@ -1181,7 +1181,8 @@ describe('a cache miss', () => {
     expect(options[0]?.ccache).toBe(setup);
     assert(result.facts);
     expect(result.facts.ccache).toEqual({ status: 'reported', hits: 176, misses: 204, hitRatePercent: 46.3 });
-    expect(h.stdout.join('\n')).toMatch(/^ {2}compilation cache 176 hits \/ 204 misses \(46\.3%\)$/m);
+    expect(h.stderr).toContain(phaseLine('cache', 'compilation cache 176 hits / 204 misses (46.3%)'));
+    expect(h.stdout.join('\n')).not.toContain('compilation cache');
   });
 
   test('builds only the ABI selected for an owned emulator and scopes the cache key', async () => {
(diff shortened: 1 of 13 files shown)


janicduplessis:
Two independent in-session reviews are clear at 25ad2774aad5477c5da18fe59e70b5b3c510a3d2. They checked engine failure returns, command failure paths, human and JSON output, launch plumbing, tests, and guides. No actionable findings; 14 focused regressions also passed during review.

Local validation passed: 3,897 unit tests, one skipped; 78 end-to-end tests; formatting, lint, build, typecheck, knip, and runtime checks. Node 22, Node 24, and published Node 20.19.4 runtime CI are green.

PR #706, opened 2026-09-11
test: reduce benchmark workflow coaching

### LINKED ISSUE #705: Reduce benchmark workflow coaching for future campaigns
## Problem
Benchmark prompts prescribe implementation and diagnostic steps rather than just tasks. Both arms require a separate log query even when launch output already proves the error, and crash prompts reveal the failure category and location.

## Evidence
scripts/agent-benchmark/driver.mjs promptFor and launch-crash-setup.mjs give warm/start/copy/poll instructions and an exact eight-command proof protocol. User requests minimal, fair prompts for the next run; current published results must remain unchanged.

## Cause
Collection and export validators depend on prescribed command forms, so shortening prompts alone rejects otherwise valid runs.

## Fix idea
Keep task goals and shared isolation/proof constraints; remove workflow coaching and crash hints. Accept actionable initial launch output without a redundant log read, and judge proof semantically rather than prescribing shell sequencing. Preserve pinned runtime, owned devices, timed worktree setup and historical evidence.

janicduplessis:
## Description

Future benchmark runs should test how agents solve the task, not how closely they follow our launch recipe. The current prompts reveal the crash category/location and prescribe warm, start, log polling and dependency-copy steps.

Fixes #705.

## Solution

Give both arms the same outcome-focused task, with only tool choice and owned-device allocation differing. Keep worktree setup timed and leave normal Stim usage to its installed skill. The shared screenshot/recording protocol remains explicit for collection.

No published data or timings change: recorded prompt policy keeps old runs on their original audit contract. New runs accept an actionable initial launch response instead of forcing another log query, replacing the [deliberate separate-capture gate](https://github.com/appandflow/stim/commit/14ac4ea4ffc695af3b45573829fc50a28a07f5f5).

## Test plan

- Exercise both arms on iOS/Android with actionable launch output, missing context, failed launch status, and earlier separate runtime logs; confirm only genuine runtime evidence allows source inspection.
- Run driver prompt/isolation/Android self-tests and compare generated task text across arms. No model runs were dispatched; the next campaign needs a pilot under the shortened instructions.

### DIFF
--- a/docs/agent-benchmark.md
+++ b/docs/agent-benchmark.md
@@ -381,8 +381,8 @@ Inject a deterministic root-render JavaScript exception before the first app
 screen, then give each agent the same repair task. Compare first recorded activity to the
 first actionable diagnosis, commands and tokens to diagnosis, and first recorded activity to
 a repaired Settings screenshot. Run separate iOS and Android blocks. The Stim
-arm must preserve the matching `stim ios` or `stim android` launch
-output and `stim logs --errors`; control collects the equivalent Metro and
+arm uses the matching `stim ios` or `stim android` launch
+output or `stim logs --errors`; control collects the equivalent Metro and
 device logs manually, using the exact emulator serial for Android `adb`
 commands. The injected error text is unique per run so the
 collector can prove that the reported stack and repair refer to this failure.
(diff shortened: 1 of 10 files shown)


janicduplessis:
Fresh independent review is clear on 71377f2938d7cc0ad2705c0012ae4ea2d8a10f4f. Resolved the prompt/validator mismatches for dependency installs and runtime-error capture. All local checks passed (4,066 tests, 79 end-to-end tests), plus the three benchmark driver self-tests. No model runs were dispatched and no published benchmark data changed. Final-head CI is running.

PR #711, opened 2026-09-11
fix: identify source paths and wait times in warm output

### LINKED ISSUE #708: warm phase lines do not name their subject, and two agents misread them the same way
Found by a deliberate two-agent concurrency test of `warm --refresh` on a source checkout 42 commits behind its upstream. The mechanism worked correctly -- one refresh, one install, the second caller waited 40s and inherited the result, no leaked claims. What did not work is the reading. Two independent agents each misread a different phase line, both because the line does not say what it is about.

## 1. The wait line attributes the waiter's elapsed time to the holder

```
  lock        waiting on stim worktree warm (pid 72423, 30s elapsed)
  lock        acquired (waited 40s for stim worktree warm pid 72423)
```

`engine/warm-claim.ts:101` takes `elapsedMs` and prints it inside the parenthesis that belongs to `pid`:

```ts
function waitingLine(holder: WarmClaimHolder, elapsedMs: number): string {
```

The call site at `:325-329` passes `now() - started`, where `started` is **the waiter's** start. So `30s elapsed` means "I have been waiting 30s", but it reads as "pid 72423 has held this for 30s".

The agent that hit it concluded the holder had held for 30s, then reported the next line as contradictory, since it appeared to have waited 40s -- longer than the holder had been holding. Its words: "the two numbers in the lock lines are measured on different clocks ... B appears to have waited *longer* than the holder had been holding."

Suggested shape, which puts the number on its real subject:

```
  lock        waiting 30s for stim worktree warm (pid 72423)
```

## 2. The deps line does not say which tree it installed in

```
  deps        pnpm-lock.yaml changed -> pnpm install (9s)
  carry       copied .DS_Store, .claude, .expo, +57 from /Users/janicduplessis/Developer/tlon-apps
```

`--refresh` installs in the **source checkout**, then `carry` copies into the worktree. The `deps` line names neither, and `carry` is the only line carrying a path. The other agent read the order as the worktree installing first and then being copied over, and flagged it as causally backwards: "a 9-second install of a 192k-file `node_modules` in an empty worktree only makes sense if the content was already materialized."

Nothing is wrong with the order. The line is just silent about where the work happened, and `carry`'s trailing `from <path>` is what makes the ambiguity land.

## Why this is worth fixing rather than filing as cosmetic

These lines exist to be read by an agent deciding whether to continue or abort, and the first thing they produced in a real two-agent run was two wrong conclusions about who was doing what. Neither agent was careless -- both quoted the output accurately and reasoned from it.

## Lower-priority observations from the same run

- There is no `lock released` line, so a log gives no way to see when the claim was dropped. For a waiter, the holder's exit is invisible.
- `carry copied .DS_Store, .claude, .expo, +57` leads with `.DS_Store`. Faithful -- it copies ignored entries, and that is the contract -- but it reads oddly as the headline item.
- `pods ... ios/Podfile.lock unchanged -> skipped` sits next to a worktree that does have `ios/Pods`, because `carry` brought it. One agent flagged that "anyone reading Pods: present as pod install ran would be wrong." That one is the design working, but it is the same shape of ambiguity.

## Evidence

Both runs exited 0. A refreshed in 50s; B waited 40s and every refresh step was a no-op (`checkout develop up to date`, `deps ... matches the last completed install -> skipped`). Afterwards: source checkout at `5fa5985329` and 0 behind, both worktrees at that sha with `node_modules`, `ios/Pods` and `.env.local`, exactly one `warm-installs` ledger entry with `completed: true`, and no claims left in `warm-locks`.

janicduplessis:
## Description

`worktree warm` places the caller's wait duration beside the holder's PID, making it look like the holder's elapsed time. Refresh install and skip messages also omit their source paths, so they can read as work performed in the destination worktree.

## Solution

Print `waiting 30s for stim worktree warm --refresh (pid 41233)`. Prefix dependency and Pods results with their source directory, using the actual lockfile directory for dependencies, including nested app lockfiles. Update the guide examples.

## Test plan

- The wait regression checks two progress intervals and the caller's total wait.
- Real Git fixture tests exercise installs at the repository root and a nested app root, skipped installs, and the Pods source path. All five targeted regressions fail on the old output and pass with this change; command arguments and working directories remain unchanged.

Fixes #708

### DIFF
--- a/packages/stim-cli/src/__tests__/warm-claim.test.ts
+++ b/packages/stim-cli/src/__tests__/warm-claim.test.ts
@@ -188,23 +188,27 @@ test('a refresh that gives up while copies drain leaves no claim of its own behi
   after.release();
 });
 
-test('a wait prints its holder every progress interval, in the shape build waits use', async () => {
+test('a wait attributes elapsed time to the waiter and names its holder every progress interval', async () => {
   const refresh = await acquireWarmClaim({ repositoryRoot: REPO, phase: 'refresh', sleep: never });
   const lines: string[] = [];
-  let clock = 0;
+  let clock = 120_000;
   let polls = 0;
   const copy = await acquireWarmClaim({
     repositoryRoot: REPO,
     phase: 'copy',
-    now: () => (clock += 10_000),
+    now: () => clock,
     progressMs: 30_000,
     out: (line) => lines.push(line),
     sleep: async () => {
+      clock += 10_000;
       if (++polls === 8) refresh.release();
     },
   });
-  expect(lines[0]).toMatch(/^ {2}lock {8}waiting on stim worktree warm --refresh \(pid \d+, \d+m?\d*s elapsed\)$/);
-  expect(lines.length).toBeGreaterThan(1);
+  expect(lines).toEqual([
+    `  lock        waiting 30s for stim worktree warm --refresh (pid ${process.pid})`,
+    `  lock        waiting 1m00s for stim worktree warm --refresh (pid ${process.pid})`,
+  ]);
+  expect(copy.wait.waitedMs).toBe(80_000);
   copy.release();
 });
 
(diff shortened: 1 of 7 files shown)


janicduplessis:
Two independent in-session reviews cleared head 28dff5fc037e5f2b36c9ad1beac9222e0a3bba59. Focused output and guide regressions passed; the full local unit and e2e suites passed. Required CI is green at this head.

Issue #696, opened 2026-09-11
Three narrow edges in the warm claim, and the degraded copy's residual race


janicduplessis:
Follow-ups from the review of appandflow/stim#690, kept out of that PR deliberately: each is narrow, none blocks the release, and #690 documents them.

## 1. A degraded copy can still overlap a refresh that starts after it

Plain `warm` degrades when no claim can be recorded (an unwritable `STIM_HOME`, or `unique-pid` unavailable). Before copying it surveys the claim set and refuses if a refresh holds it, which closes the common case. It cannot close the case where the refresh starts *after* the copy is already running: the copy holds nothing, so there is nothing for the refresh to wait on.

Reachability is limited by the fact that the degrade and the refresh need different answers from the same filesystem -- one process unable to write `STIM_HOME` while another can. Two agents under different sandbox policies on one machine is the realistic shape.

A read cannot fix this; only a claim can, and in that state no claim can be recorded. Options worth pricing: a claim record that does not require process identity (weaker liveness, but enough to be seen), or having the copy verify `node_modules` against the lockfile afterwards and refuse the worktree on a mismatch -- detection instead of prevention, which fails loudly rather than silently.

## 2. `$STIM_HOME/warm-locks` as a plain file refuses forever, with a remedy that cannot work

    printf x > $STIM_HOME/warm-locks
    stim worktree warm
      Could not refresh ...: the claim path is a file, not a claim directory
      rm -rf '<home>/warm-locks/<name>.lock'
      failed: STIM_CLAIM_REFUSED

The offending path is `warm-locks`; the refusal names `warm-locks/<name>.lock`, which does not exist, so running the printed remedy changes nothing and the next run refuses identically. `origin/main` copies fine in the same fixture.

Root cause is in the primitive rather than in warm: `tryAcquireClaim`'s `mkdirSync` `ENOTDIR` calls `refuse(root, root, ...)`, naming the claim-set directory instead of the non-directory ancestor it actually hit. Realism is low -- nothing Stim ships ever writes `warm-locks` or `$STIM_HOME` as a file -- but the remedy being a no-op makes it unrecoverable without guessing.

## 3. One refusal carries no code and no remedy

`--refresh` against a mode-`0500` `STIM_HOME` prints only:

    Could not warm this worktree: EACCES: permission denied, mkdir '<home>/warm-locks/<name>.lock'

[...]

PR #713, opened 2026-09-11
fix: let plain warm copy when its claim store is read-only

### LINKED ISSUE #704: A read-only STIM_HOME makes plain warm refuse, naming a contender that never existed
## What happens

With `STIM_HOME` on a read-only mount (or with a dangling symlink anywhere in its path), `tryAcquireClaim` refuses:

```
ClaimRefusedError | code=STIM_CLAIM_REFUSED
reason=another process took and gave up the claim on every one of 64 attempts to take it
```

There was no other process. Plain `warm` then refuses the copy, because `warmClaimDegradation` degrades a refusal only when its reason is `CLAIM_PATH_NOT_A_DIRECTORY`.

This is a regression against 1.0.0 for that user: before the claim existed in `warm`, an unwritable `STIM_HOME` did not stop a copy that only reads.

## Why

`mkdirSync(root, { recursive: true })` against a read-only mount reports `ENOENT` on macOS -- node's `MKDirpSync` returns the errno of the `stat` it takes after the failed mkdir, not the mkdir's own. `ENOENT` is in `contended()` (`ownership-claim.ts:442`), because it is also what a claim set another process just removed gives. So the acquisition loop `continue`s through all 64 attempts and reports exhaustion as contention (`ownership-claim.ts:543`).

## Scope

Narrow, and the neighbouring cases are fine:

- `EACCES` (a `chmod 555` parent) is not in `contended()` and carries no `STIM_` code, so `warmClaimDegradation`'s last branch degrades it correctly. Verified.
- A path holding a plain file gives `ENOTDIR`/`EEXIST` and refuses with `CLAIM_PATH_NOT_A_DIRECTORY`, which degrades. Verified.

Only the read-only *mount* / broken-symlink `ENOENT` lands wrong.

## Reproduction

No read-only mount needed -- a dangling symlink in the path gives the same persistent `ENOENT`:

```js
import { tryAcquireClaim } from './packages/stim-cli/src/ownership-claim.ts';
// ln -sfn /nonexistent-target-xyz ro-link
tryAcquireClaim({ root: `${dir}/ro-link/store`, mode: 'exclusive', label: 'probe' });
```

## History

#681 shipped a fix for exactly this: `createClaimDir` walked up so the kernel's own errno for the failing level survived, `provesAContender` distinguished `EEXIST`/`ENOTEMPTY` from `ENOENT`/`EINVAL`, and exhaustion without proof of a contender became a `ClaimStorageError`. #690 replaced that with the narrower `CLAIM_PATH_NOT_A_DIRECTORY` degradation trigger and removed the machinery, along with `packages/stim-cli/src/__tests__/ownership-claim-storage.test.ts` (-92 lines).

Restoring the removed test at `58b941542` gives 3 failures, so the deleted test is still an accurate statement of the desired behaviour.

## Also stale

`engine/warm-claim.ts:120-127` still promises that "a claim store the filesystem rejects" degrades. True for `EACCES`, false for this path.

## Shipping

Documented as a known limitation in the 1.1.0 notes rather than fixed, because the fix means reversing a deliberate design decision in #690 and would want its own review.

janicduplessis:
## Description

Plain `worktree warm` refuses a copy when its claim directory is on a read-only mount or below a dangling symlink, incorrectly reporting 64 collisions with another process. Node can report `ENOENT` for recursive mkdir on a read-only filesystem ([upstream issue](https://github.com/nodejs/node/issues/47098)), which the claim treats as publication contention.

## Solution

Recover the direct filesystem error when recursive mkdir fails, and distinguish a proven dangling symlink from a directory removed during publication. Plain warm can then use its existing storage-failure fallback; `--refresh` still refuses before changing the source checkout. Unresolved claims and repeated publication races still refuse. Cleanup reports a storage failure for the affected claim and continues the sweep.

Restore the storage distinction removed by [the warm-refresh change](https://github.com/appandflow/stim/commit/97ede367fcd80b04773b575f8349f3e3dbe4133b). The degraded copy retains the race documented in #696: a refresh starting after its claim survey is not covered.

## Test plan

- Real dangling-symlink regressions failed before the fix; plain warm now copies with empty stdout, while `--refresh` leaves the source HEAD and destination unchanged.
- Real read-only APFS image, macOS/Node 22.22.1: `ENOENT` became `EROFS` for both claim modes with absent and existing claim directories.
- Real filesystem publication races recover when transient and remain `STIM_CLAIM_REFUSED` when repeated, including occupied-directory rename collisions.
- A read-only build lock reports a failed clearance; `gc --delete` keeps it, clears the next healthy build slot, and reports one failed entry.
- Required checks passed; full unit suite: 4,107 passed, 1 skipped. All 79 end-to-end tests passed.

Fixes #704

### DIFF
--- a/packages/stim-cli/src/__tests__/ownership-claim-publication.test.ts
+++ b/packages/stim-cli/src/__tests__/ownership-claim-publication.test.ts
@@ -1,0 +1,86 @@
+import { existsSync, mkdirSync, mkdtempSync, rmSync, writeFileSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join, relative } from 'node:path';
+import { readClaimSet, releaseClaim, tryAcquireClaim } from '../ownership-claim.ts';
+
+const faults = vi.hoisted(() => ({
+  readOnly: '',
+  beforeRename: null as null | ((from: string, to: string) => void | (() => void)),
+}));
+
+vi.mock('node:fs', async (importOriginal) => {
+  const fs = await importOriginal<typeof import('node:fs')>();
+  return {
+    ...fs,
+    mkdirSync: (...args: Parameters<typeof fs.mkdirSync>) => {
+      const path = String(args[0]);
+      if (!faults.readOnly || relative(faults.readOnly, path).startsWith('..')) return fs.mkdirSync(...args);
+      const options = args[1];
+      const recursive = typeof options === 'object' && options !== null && options.recursive === true;
+      if (recursive && fs.existsSync(path)) return undefined;
+      const code = recursive ? 'ENOENT' : 'EROFS';
+      throw Object.assign(new Error(`${code}: mkdir '${path}'`), { code });
+    },
+    renameSync: (...args: Parameters<typeof fs.renameSync>) => {
+      const cleanup = faults.beforeRename?.(String(args[0]), String(args[1]));
+      try {
+        return fs.renameSync(...args);
+      } finally {
+        cleanup?.();
+      }
+    },
+  };
+});
+
+let home: string;
+let root: string;
+
+beforeEach(() => {
+  home = mkdtempSync(join(tmpdir(), 'stim-claim-publication-'));
+  root = join(home, 'build.lock');
+  process.env.STIM_HOME = home;
+});
+
+afterEach(() => {
+  faults.readOnly = '';
+  faults.beforeRename = null;
+  delete process.env.STIM_HOME;
+  rmSync(home, { recursive: true, force: true });
+});
+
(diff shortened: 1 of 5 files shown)


janicduplessis:
Both independent reviews cleared final head b591789dc070d17352afa14e2760c8a48b732278 after the GC acquisition-error finding was corrected. Direct clearance and GC continuation regressions pass, along with 4,107 unit and 79 e2e tests and real read-only APFS checks. Required CI is green. The existing post-acquisition staging-cleanup failure is tracked separately in #716.

PR #715, opened 2026-09-11
fix: recover transient Android adoption cleanup failures

### LINKED ISSUE #639: Recover transient app-data cleanup failures after Android pool adoption
## Problem
The first launch on an adopted Android emulator can fail at adb shell pm clear even though the exact owned AVD and target app are identified. A later identical clear succeeds, and retrying Stim launches successfully.

## Evidence
Observed with published stim1.0.0, Android Emulator36.4.9/API36 on macOS26.5.2. Emulator output says snapshot default_boot loaded in3683ms; Stim reports prepared6.4s and artifact fingerprint hit, then STIM_LAUNCH_FAILED for pm clear. The same clear succeeds about14s later; the next Stim call reuses the APK and reaches readiness. Emulator reports boot completed11754ms; retained device logs show system services restarting and boot reason reboot.

## Cause
Not fully established. A transient post-snapshot reboot/ADB disconnect is consistent with the evidence. resetAdoptedAvd currently fails immediately on any clear/uninstall command error, and the final error drops useful child-process diagnostics. A successful snapshot-load marker alone does not prove services remained ready.

## Fix idea
Reproduce with an owned disposable AVD and record boot identity plus exact failure output. Consider bounded recovery on idempotent adoption cleanup failures, rechecking owned AVD identity before every destructive retry and waiting for device readiness. Keep adoption pending on failure and never launch without successful isolation cleanup. Avoid adding delays to successful adoption.

janicduplessis:
## Description

Adopting a parked Android emulator can fail with `adb: device offline` or `error: closed` while listing or clearing apps, even when retrying later succeeds. The original snapshot failure's cause remains unknown; this change covers recovery during adoption cleanup, not later install or launch failures.

## Solution

Retry cleanup within a 30-second recovery budget, including its ownership and boot-readiness probes. Each attempt reads the installed packages again and verifies the same owned AVD before clearing or uninstalling an app. Successful cleanup adds no wait; other errors still block launch, leave adoption pending, and preserve the child-process diagnostics.

## Test plan

- Regression cases cover interrupted package listing, clearing, and uninstalling; an unavailable or replaced AVD; exhausted recovery; and a permanent cleanup error that preserves diagnostics and prevents launch.
- On Emulator 35.6.11/API 36, a guest reboot induced a real `adb: device offline`; cleanup recovered in 19.2 seconds, cleared retained-app data, and removed another third-party app. Normal cleanup took 244 ms. The disposable AVD was removed afterward.

Fixes #639

### DIFF
--- a/packages/stim-cli/src/__tests__/android-command.test.ts
+++ b/packages/stim-cli/src/__tests__/android-command.test.ts
@@ -471,7 +471,7 @@ test('an invalid Android pool bound refuses before device creation and emits one
 describe('adopted Android installs', () => {
   afterEach(() => resetExecutor());
 
-  test.each(['same', 'different', 'signature', 'downgrade', 'cleanup-failed', 'install-failed'] as const)(
+  test.each(['same', 'different', 'signature', 'downgrade', 'cleanup-failed', 'error', 'install-failed'] as const)(
     'cleanup precedes install and only matching APK bytes skip installation: %s',
     async (mode) => {
       const apkPath = fakeApk();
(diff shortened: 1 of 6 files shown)


janicduplessis:
Two independent in-session reviews cleared head 530e2c07436e96a839cf241d59ffd6934cbe7ed3. Focused recovery, ownership, install and readiness regressions pass. Required CI is green. Real fault injection on an owned disposable AVD recovered after adb device offline, verified cleared app data and removed the other app, and completed centralized teardown.

PR #718, opened 2026-09-11
fix: release cleanup claims when staging removal fails

### LINKED ISSUE #716: Contain abandoned claim staging cleanup failures and release the cleanup claim
## Problem

A filesystem error while garbage collection removes abandoned claim staging files escapes `clearFreeClaimSet`, aborts the remaining sweep, and leaves the cleanup operation's own exclusive claim live until its process exits.

## Evidence

Reproduced against origin/main code at 2efb9db842313aa7dcb0ab6d2bc60fe052a8ad22 on macOS with real filesystem permissions:

1. Create a disposable claim root containing `.staging-interrupted/partial.claim`.
2. Set that staging directory to mode 000 while leaving the claim root writable.
3. Call `clearFreeClaimSet({ root })`.
4. It throws EACCES. `readClaimSet(root)` reports one live exclusive claim owned by the calling process.

Restoring permissions and explicitly releasing the caller's claim allows cleanup. The reproduction removed all its temporary files afterward.

## Cause

`removeAbandonedStaging(root)` executes after the cleanup claim is acquired but before `releaseClaim`, without a `finally`. Its `rmSync` errors are not contained. The GC build-lock and build-slot loops call this helper without a surrounding per-entry catch.

This is separate from #704 / #713's acquisition-time read-only storage classification. It exists on the base branch and needs post-acquisition cleanup handling.

## Fix idea

Release the cleanup claim in a finally block, report the staging deletion failure through the existing failed-clearance result, and keep unrelated GC entries eligible for cleanup. Preserve undeletable staging files for a later retry. Add observable regressions for claim release and continuation to another resource.

janicduplessis:
## Description

A permission error while `gc --delete` removes abandoned claim staging files aborts the sweep and leaves its own exclusive claim held until the process exits.

## Solution

Report the staging error through the existing failed-clearance result and release the cleanup claim in `finally`. Undeletable staging files remain for a later retry, while other cleanup entries can still be reclaimed.

## Test plan

- Real mode-000 staging directory: failed before the fix; now reports `EACCES`, leaves no live cleanup claim, preserves the payload, and clears successfully after permissions are restored.
- `gc --delete` reports the blocked build lock, clears a later healthy build slot, and counts one failed entry.

Fixes #716

### DIFF
--- a/packages/stim-cli/src/__tests__/gc.test.ts
+++ b/packages/stim-cli/src/__tests__/gc.test.ts
@@ -2717,6 +2717,32 @@ test('--delete removes the stale lock and leaves the live one alone', async () =
   expect(output).toMatch(/build lock/i);
 });
 
+test.skipIf(process.getuid?.() === 0)(
+  '--delete continues after staging cleanup fails and leaves no cleanup claim',
+  async () => {
+    saveConfig({ version: 2, projects: {}, repos: {} });
+    installExecutor();
+    const blocked = join(tmpHome, 'build-locks', 'ios-staging.lock');
+    const staging = join(blocked, '.staging-interrupted');
+    const writable = join(tmpHome, 'build-slots', 'slot-0');
+    mkdirSync(staging, { recursive: true });
+    mkdirSync(writable, { recursive: true });
+    writeFileSync(join(staging, 'partial.claim'), '{');
+    chmodSync(staging, 0);
+    try {
+      const output = await captureLog(() => sweepingGc({ delete: true }));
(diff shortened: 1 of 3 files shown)


janicduplessis:
Two independent in-session reviews cleared head 52d7a9617fbb1a11eb7e261d3a161ed78ed83961. Real permission-failure regressions verify claim release, preserved staging, retry after permissions recover, and GC continuation. Full local validation passed with 4,120 unit tests and 79 e2e tests. Required CI is green at this head.

PR #726, opened 2026-09-11
docs: route agent situations to specific guide sections

### LINKED ISSUE #723: guide agent: replace the run-on routing paragraph with a situation-to-section table, and move the timeout bullet next to ios/android
## What happened

Across two agent sessions on stim 1.1.0, each ran `stim guide agent` exactly once and never opened a sub-topic, including while waiting nine minutes on another workspace's build and after a `stim ios` call outlived a 10-minute tool timeout (see the sibling issue on the phase line).

## Why

`LOAD ADVANCED GUIDANCE WHEN NEEDED` is 21 topic lines followed by one sentence listing about fourteen situations ("...cache misses, bypasses, or concurrent builds; capacity limits; cache statistics from stim stats; worktree carry-over; fingerprint exclusions; gc; --force; cleanup failures; or unfamiliar states and error codes"). There is no situation-to-section mapping an agent can match against what it is seeing, so the list reads as a table of contents rather than a routing rule.

The one bullet that applied ("A native build can outlive a shell timeout. Retry the same command ...") sits in the middle of the general rules, far from the `stim ios` / `stim android` commands it is about.

## Proposal

1. Replace the paragraph with a table: left column is the situation as the agent experiences it, right column is the exact `stim guide <topic> <section>` to read. First rows:
   - a build is `waiting on` another workspace, or a tool call timed out during a build -> `lifecycle concurrency`
   - a `--variant` or scheme question, or `assembleDebug` produced several APKs -> `lifecycle options`
   - a refusal with a CODE -> `errors <CODE>`
   - running under a sandbox -> `errors sandbox`
2. Move the timeout and shared-build bullet to sit directly under the `stim ios` / `stim android` command block, since that is the moment it applies.
3. Keep the full topic list, but after the table.

janicduplessis:
## Description

Advanced situations share one paragraph after the topic list, without pairing what an agent sees with the section to read. Native timeout advice also sits away from the platform commands.

## Solution

Put a situation-to-command table before the full topic list and move timeout and shared-build guidance directly below `stim ios` / `stim android`. Retry advice applies only when the tool call timed out. The routes use existing guide topics and sections; the static skill stays unchanged.

## Test plan

- All 37 guide tests passed, including validation of every table route and the full topic list.
- Ran all 15 table commands through the built CLI with successful, nonempty output, substituting `STIM_NO_METRO` for `<CODE>`.

Fixes #723

### DIFF
--- a/packages/stim-cli/src/__tests__/guide.test.ts
+++ b/packages/stim-cli/src/__tests__/guide.test.ts
@@ -376,11 +376,28 @@ test('the agent and lifecycle guides name both workflows', () => {
   }
 });
 
-test('the agent guide routes to every detailed topic', () => {
+test('the agent guide routes situations to valid sections before listing every topic', () => {
   const agent = renderTopic('agent');
   assert(agent);
+  const [situations, topicList] = agent.split('FULL TOPIC LIST');
+  assert(situations);
+  assert(topicList);
+  const rows = situations
+    .split('\n')
+    .filter((line) => line.startsWith('| '))
+    .slice(2);
+  expect(rows.length).toBeGreaterThan(0);
+  for (const row of rows) {
+    const route = /^stim guide ([a-z]+)(?: ([\w<>]+))?$/.exec(row.split('|')[2]?.trim() ?? '');
+    assert(route, row);
+    const [, topic, section] = route;
+    assert(topic);
+    expect(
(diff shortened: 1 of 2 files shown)


janicduplessis:
Two independent in-session reviews cleared head 14d79a19798b2b203e4ba3c70542f775a75f54f8. Every table route resolves, full topic routing and safety instructions are retained, and timeout retry advice is conditional. All required local checks and CI passed.

PR #753, opened 2026-09-11
fix: detect stalled iOS simulators and report memory pressure

### LINKED ISSUE #752: Bound iOS simulator startup waits and diagnose host memory pressure
## Problem

A local iOS simulator can report Booted while processes inside it cannot start. Stim then accepts readiness and can wait indefinitely in simulator launch operations. Failures point at the app without mentioning observed host memory pressure. SimSlim is already supported, but doctor only reports a missing executable after a profile is configured; it does not recommend the existing integration for parallel simulator work.

## Evidence

A user reported successful bootstatus, installation, and screenshots while appinfo, launch, terminate, and openurl timed out. Process samples showed simulator daemon spawns stalled in dyld. Host memory compression, swap use, and load were high. Memory pressure is a plausible cause, but freeing memory was not tested, so OOM is not established.

On origin/main ff786f1e6, launchIosApp in packages/stim-cli/src/engine/app-install.ts passes no timeout to defaults write or openurl. ensureIosBooted in engine/device.ts accepts Booted or a fulfilled boot promise without an in-simulator spawn probe. doctor.ts checkSimSlim returns no finding when no profile is configured.

## Cause

Unbounded tool calls and shallow readiness checks are confirmed in Stim. The underlying CoreSimulator stall is not diagnosed. A timeout or a low free-memory figure alone must not be labeled an OOM crash.

## Fix idea

Bound the affected simulator calls and verify process spawning before accepting local simulator readiness. On a readiness or launch failure, report observable macOS memory pressure when available and recommend freeing memory before retrying. Give doctor a read-only memory-pressure diagnostic and recommend the existing optional SimSlim profile integration, with its service tradeoffs. Keep unknown pressure unknown, preserve structured error contracts and ownership, and never automatically restart other workspaces, kill applications, install SimSlim, or enable a profile.

Validate timeout/refusal and normal paths with targeted tests, real bounded simulator commands on a fresh owned device, and read-only macOS pressure queries. Do not deliberately exhaust host memory.

janicduplessis:
## Description

A simulator can report Booted while processes inside it cannot start. Stim accepted that state and could wait indefinitely in dev-client preparation or `openurl`. The reported incident coincided with severe host memory use; its underlying cause was not confirmed.

Fixes #752.

## Solution

Require a process-spawn probe to complete within 30 seconds before installation, and bound the missing simulator preparation and launch calls to 60 seconds. Report macOS memory pressure before boot and on startup failures when available; unknown readings remain unknown, and timeouts are not labeled OOM crashes. Recovery guidance puts freeing memory ahead of another launch attempt.

Doctor recommends optional SimSlim profiles for parallel iOS work without installing or enabling them; review disabled services for effects on app behavior and tests.

## Test plan

- Regression cases cover unresponsive simulators, bounded calls, memory-pressure interpretation, and platform isolation.
- On macOS with iPhone 17 / iOS 26.5, the candidate installed and launched an Expo dev client on a fresh Stim-owned simulator (`launched: true`). Real `simctl spawn /usr/bin/true`, `defaults write`, and `openurl` calls succeeded; cleanup removed the exact owned simulator and worktree.
- Read-only macOS pressure query returned normal. Deliberate host memory exhaustion was not tested.

### DIFF
--- a/packages/stim-cli/README.md
+++ b/packages/stim-cli/README.md
@@ -58,6 +58,13 @@ one native platform is in scope; shared project checks still run. Doctor also
 prints the running CLI version and the `stim` installation resolved from PATH,
 and flags a resolved installation that is older than another available one.
 
+For parallel iOS work, SimSlim is recommended as an optional way to reduce
+simulator background services and memory use. Run `stim guide lifecycle simslim`
+to review the service tradeoffs and configure a profile. Doctor recommends the
+setup without installing or enabling it. When a simulator cannot start
+processes, Stim bounds the wait and reports observed host memory pressure when
+available; free memory before retrying. A timeout alone does not prove OOM.
+
 For stale Android CMake launcher findings, stop native builds and run
 `stim doctor --fix --platform android` in the affected checkout. It clears
 affected ignored, untracked generated `.cxx` configurations in the app and
--- a/packages/stim-cli/src/__tests__/doctor.test.ts
+++ b/packages/stim-cli/src/__tests__/doctor.test.ts
@@ -1212,8 +1212,8 @@ test('runDoctor emits one concurrency note when a limit is set', () => {
   rmSync(project, { recursive: true, force: true });
 });
 
-test('checkSimSlim reports only configured profiles that cannot run', () => {
-  expect(checkSimSlim()).toBeNull();
+test('checkSimSlim recommends optional profiles and reports configured profiles that cannot run', () => {
+  expect(checkSimSlim()).toMatchObject({ level: 'note', fix: expect.stringContaining('stim guide lifecycle simslim') });
   expect(checkSimSlim({ configured: true, onPath: true })).toBeNull();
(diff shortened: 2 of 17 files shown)


janicduplessis:
Two fresh in-session reviewers reviewed the issue, full diff, tests, and guidance. One found that the new host-memory remedy also appeared for remote launch failures; the final change limits it to local simulators and adds a command-level regression for both paths. Both reviewers cleared the correction. No remaining actionable findings.

Final local checks passed: formatting, lint, build, typecheck, 4,219 tests (one existing skip), and 84 end-to-end tests. The website build and unused-code check also passed. A fresh owned iOS 26.5 simulator passed normal Expo launch and real bounded spawn/defaults/openurl calls, then was removed with its worktree. Host memory exhaustion was not induced. CI must pass on ac5f8c090 before merge.

Separate pre-existing timeout boundaries are tracked in #754.

PR #756, opened 2026-09-12
fix: allow cold Android emulators more time to load

### LINKED ISSUE #749: android: the 20s UNVERIFIED window is too short right after an emulator is created
## What happened

stim 1.1.0, tlon-apps, `stim android --variant productionDebug` in a fresh worktree. Stim created the emulator (about 35s), installed the APK from the shared cache, launched, and reported:

```
UNVERIFIED: no bundle request reached this workspace's Metro ... within 20s
```

The bundle request reached Metro about 3 seconds after the deadline and bundled normally; the app was fine. The run's summary still said UNVERIFIED, so the agent treated launch as suspect and spent time checking a healthy app.

Seen once, on an M4 Pro under load from a sibling worktree's xcodebuild, but a first launch on a just-created emulator is slow on any machine: the AVD is cold, the app's first start does its own initialisation, and only then does the dev client ask Metro for a bundle.

## Proposal

Scale the verification window to the situation: a launch on an emulator Stim created in this same run (or booted less than a minute ago) gets a longer budget, say 60s, and the summary says which budget applied. The 20s default is fine for a warm emulator.

janicduplessis:
## Description

A healthy app on a newly created Android emulator can request its first bundle just after Stim's 20-second verification window, leaving an unverified result even though the app loads normally.

Fixes #749.

## Solution

Allow up to 60 seconds for the first bundle on a new owned Android emulator and print the selected budget. Existing emulators retain the 20-second window. Fatal errors still end verification immediately, and remote, physical-device, and release behavior is unchanged.

## Test plan

- Command regression uses the actual verifier with a bundle arriving at 23 seconds: the new-emulator run verifies, while the existing-emulator run remains unverified at 20 seconds.
- Missing or unfinished cold bundles stop at 60 seconds; a fatal error at 23 seconds stops immediately.
- Real Expo 57 / Android API 36 run printed the 60-second budget and returned unverified at that deadline while showing the dev-client picker. Resending the printed deep link loaded the expected screen with no captured errors. The late-bundle success case is covered by the controlled regression.

### DIFF
--- a/packages/stim-cli/src/__tests__/android-command.test.ts
+++ b/packages/stim-cli/src/__tests__/android-command.test.ts
@@ -63,6 +63,7 @@ import {
   installAndroidApp,
   deviceShellArg,
   androidDevClientUrl,
+  verifyLaunch,
 } from '../engine/app-install.ts';
 import type { AssetManifest } from '../engine/asset-manifest.ts';
 import { PREBUILD_ERROR } from '../engine/prebuild.ts';
@@ -204,6 +205,7 @@ interface VerifyArgs {
   since?: string | number;
   metroPort?: string | number | null;
   platform?: string | null;
+  timeoutMs?: number;
 }
 
 interface Calls {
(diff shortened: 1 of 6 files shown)


janicduplessis:
Two fresh reviewers cleared the final change. Review caught recovered AVDs incorrectly carrying the new-creation flag; recovery now keeps the warm budget, with live/offline and stopped recovery regressions. Full final checks pass: 4,225 tests (one existing skip), 84 e2e tests, format/lint/build/typecheck. Runtime, knip, and website checks passed for this change too.

Native QA verified the selected 60-second budget and truthful unverified result at the Expo picker; the printed deep-link remedy recovered the expected screen and clean error logs. This is not presented as a successful cold first launch. The QA resources were subsequently removed after the separate restart proof in #757.

PR #757, opened 2026-09-12
fix: reconnect Android emulators after serial changes

### LINKED ISSUE #750: android: an emulator that restarts under load keeps its owned status while the adb reverse and Metro link point at the old serial
## What happened

stim 1.1.0, tlon-apps, three worktrees running at once. Under a host load spike (load average around 350 during a sibling's xcodebuild) one worktree's emulator restarted on its own: serial `emulator-5564` became `emulator-5566`. `stim status` kept showing the device as owned and healthy, but:

- the adb reverse for the Metro port was set up on the old serial, so the app could not reach Metro after the restart;
- agent-device's session was pinned to the old serial, and every command failed with `device 'emulator-5564' not found` until a new session was opened.

The agent found out from agent-device's error, not from Stim. It recovered with a force-stop and the deep-link remedy Stim prints for a failed first bundle, but only after working out what had happened.

The same emulator later ANR'd under `lowmemorykiller` pressure (the AVD had 2 GB).

## Proposal

- `stim status` and the platform commands resolve the owned AVD to its current serial (`adb devices -l` reports the AVD name per serial) instead of trusting the serial recorded at boot, and say so when it changed: `emulator restarted since boot (was emulator-5564, now emulator-5566); re-establishing adb reverse`.
- Re-run the adb reverse on that path, since it is per serial.
- Optionally, a note in `stim doctor` when the AVD's RAM is small enough that the app is a likely `lowmemorykiller` target under parallel builds.

janicduplessis:
## Description

An Android emulator can restart on another serial while Stim builds the app. Installation then uses the serial captured before the build. Status also reports recorded ownership without checking the current AVD serial, making the broken connection difficult to diagnose.

Fixes #750.

## Solution

Resolve the owned AVD again before installation. A changed serial must finish booting and still identify the same AVD; the workspace must still own that assignment. Refuse when identity or readiness cannot be confirmed. Persist the new console port and use the verified serial for the rest of the run.

Status reports the detected serial and warns about changes without changing state. Recovery guidance tells users to rerun their Android command with the same build options and reopen automation on the reported serial.

## Test plan

- Command regressions cover a serial change during the build, unchanged serials, missing or unready devices, changed-again identity, and reassigned ownership. Physical and remote paths do not run local AVD probes.
- Status regressions cover changed, absent, missing, and unreadable devices while asserting the registry remains unchanged.
- Real API 36 QA emulator restarted from emulator-5554 to emulator-5556. Status reported the change without writing the registry. The candidate Android run updated the assignment, restored forwarding (verified with adb reverse --list), and launched successfully; UI and error logs were checked. The exact QA device and worktree were removed afterward.

### DIFF
--- a/packages/stim-cli/src/__tests__/android-command.test.ts
+++ b/packages/stim-cli/src/__tests__/android-command.test.ts
@@ -6,6 +6,7 @@ import assert from 'node:assert';
 import { captureProcessToken } from '../process-identity.ts';
 import { ClaimRefusedError, ClaimUnavailableError, claimRemoveCommand } from '../ownership-claim.ts';
 import { AvdRecoveryError } from '../engine/device.ts';
+import { ensureRemoteBootOwned } from '../engine/device-remote.ts';
 import { once } from 'node:events';
 import { type ChildProcess, spawn } from 'node:child_process';
 import {
@@ -22,7 +23,7 @@ import { tmpdir } from 'node:os';
 import { join } from 'node:path';
 import { Command } from 'commander';
 import { collectorProcessTitle } from '../collector/ownership.ts';
-import { loadConfig, setProjectSetting, upsertProject } from '../config.ts';
+import { loadConfig, setDevice, setProjectSetting, upsertProject } from '../config.ts';
 import { parseNdjsonText } from '../ndjson.ts';
 import { emulatorLogFile, workspaceLogsDir, workspaceStateFile } from '../paths.ts';
 import { writeWorkspaceState } from '../supervisor/run.ts';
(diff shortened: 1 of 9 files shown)


janicduplessis:
Both fresh reviewers cleared the complete issue, diff, tests and guidance with no actionable findings. Full local checks passed: 4,232 tests (one existing skip), 84 e2e tests, format/lint/build/typecheck/knip.

Real controlled Android API 36 proof moved the same private QA-owned AVD from emulator-5554 to emulator-5556. Status reported the new serial without changing registry bytes. The candidate then used the new serial, persisted the assignment, restored Metro forwarding (confirmed by adb reverse --list), and verified launch in 4.567 seconds. Expected UI and empty error logs were confirmed. Exact device, worktree, automation sessions, Metro and registry cleanup passed; the source fixture remained unchanged.

janicduplessis:
Integrated with current main and reran all checks at 5a87d836c0e05403013fa3fb485369c65a49412f. Fresh integration review is clear: cold-start timeout selection, restart identity guards, and both guide sections are preserved. Format, lint, build, typecheck, knip, runtime, 4,237 unit tests (one existing skip), and 84 end-to-end tests passed.

PR #758, opened 2026-09-12
fix: bound local iOS discovery and device operations

### LINKED ISSUE #754: Bound remaining iOS discovery, boot, install, and SimSlim child waits
## Problem

The local iOS workflow still has native operations without a caller-supplied deadline. A stuck CoreSimulator service or SimSlim discovery can therefore prevent the command from reaching the bounded readiness/launch checks in #753. This is a source-confirmed cancellation gap, not a reproduced hang in those operations.

## Evidence

At main ff786f1e68391ac5b6fafc40292af57bc5d86e1b (unchanged by #753):

- packages/stim-cli/src/sim/ios.ts: listAllIosSims defaults timeoutMs to undefined; bootIosSim runs the initial simctl boot before starting its bootstatus deadline, without an exec timeout.
- packages/stim-cli/src/engine/app-install.ts: installIosApp invokes simctl install without timeoutMs.
- packages/stim-cli/src/engine/simslim.ts: reconcileSimSlim awaits waitForChild without an outer deadline. SimSlim 0.8's apply/reboot has a ten-minute context, but its initial FindDevice/list runs before that context: https://github.com/MobAI-App/simslim/blob/v0.8.0/cmd/simslim/main.go .

The incident behind #752 had successful listing and installation; these additional operations were not demonstrated to hang in that incident.

## Cause

Timeouts cover selected nested operations, rather than every external operation that can prevent iOS startup from completing.

## Fix idea

Choose operation-appropriate bounds and cancellation behavior for these existing calls, retaining long cold-boot/install allowances and ownership claims until guarded operations stop. Verify process cleanup and accepted native arguments against real tools; simulate stuck child processes without exhausting host memory. Keep this separate from #753's confirmed spawn/launch gaps.

janicduplessis:
## Description

Local iOS discovery, the initial boot request, installation, and the outer SimSlim wait can block without a Stim deadline. SimSlim 0.8's own ten-minute apply deadline starts [after device discovery](https://github.com/MobAI-App/simslim/blob/09fc9cbbca35db5230e6d571a0a366fe6876266e/cmd/simslim/main.go#L754-L764), so upgrading alone does not cover the gap. The earlier host-pressure incident’s cause remains unconfirmed.

## Solution

Bound discovery to 30 seconds, boot to ten minutes including its initial request, installation to five minutes, and SimSlim to twelve minutes including discovery. Opening the Simulator app after boot is best-effort with a five-second bound. Direct simctl invocations opt into hard termination; other executor callers keep their existing defaults. A timed-out bootstatus child must exit before Stim surveys or retries.

SimSlim records its owned process group in a per-simulator claim. A deadline, Ctrl-C, or SIGTERM triggers identity-verified group termination with a ten-second confirmation bound; unconfirmed termination keeps the claim and blocks another reconciliation. An immediate CLI exit attempts verified group termination without releasing the claim. The simulator's managed-settings record survives interruption so retry can reconcile partial changes.

## Test plan

- Controlled Node processes prove hard timeout waits for actual exit, Ctrl-C/SIGTERM exit with 130/143, and an immediate CLI exit leaves no surviving child. Confirmed group cleanup precedes claim release; a surviving descendant after its leader exits retains the claim and blocks a second reconciliation.
- Regressions cover initial boot consuming the shared budget, bootstatus waiting for actual exit before survey, and an install timeout stopping before dev-client preparation with host recovery guidance.
- On one disposable iPhone 17 / iOS 26.5 simulator, real discovery, boot, install, and Expo launch passed: local cache hit, `launched: true`, healthy UI, empty error logs. Exact stop/remove cleanup completed and preserved all pre-existing simulator IDs. The bounded direct-argv Simulator opener also succeeded. No SimSlim profile was applied.

Fixes #754

### DIFF
--- a/packages/stim-cli/src/__tests__/engine-app-install.test.ts
+++ b/packages/stim-cli/src/__tests__/engine-app-install.test.ts
@@ -2611,3 +2611,21 @@ describe('launch verification after Metro prefetch', () => {
     expect(Boolean(result.fatal)).toBe(event === 'bundle_response_failed');
   });
 });
+
+test('a simulator install timeout fails before dev-client preparation and gives host recovery guidance', () => {
+  const exec = recordingExec({ fail: 'simctl install', failCode: 'ETIMEDOUT' });
+  const result = installIosApp(
+    {
+      udid: 'U1',
+      appPath: '/private/app with spaces.app',
+      bundleId: 'com.example.app',
+      devClientScheme: 'example',
+      proveInstalled: false,
+    },
+    { exec },
+  );
+  expect(result).toMatchObject({ failed: true, code: INSTALL_ERROR });
+  expect(result.reason).toMatch(/simctl install failed.*Activity Monitor/);
+  expect(exec.options[0]).toEqual({ timeoutMs: 300000, killSignal: 'SIGKILL' });
+  expect(exec.calls.some((args) => args.includes('defaults'))).toBe(false);
+});
--- a/packages/stim-cli/src/__tests__/engine-device.test.ts
+++ b/packages/stim-cli/src/__tests__/engine-device.test.ts
@@ -82,7 +82,8 @@ describe('ensureBooted: ios', () => {
         commands.push(cmd);
         return '';
       },
-      runFile: (file, args = [], options) => {
+      runFile(file, args = [], options) {
+        if (file === 'xcrun' && args[1] === 'list') return this.run!([file, ...args].join(' '));
         probes.push([file, ...args, options]);
         return '';
       },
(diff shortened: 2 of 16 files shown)


janicduplessis:
Fresh review found a cancellation regression: detaching the SimSlim process group leaves it running if Ctrl-C terminates Stim, which also removes the parent-owned deadline. A controlled Node reproduction using the actual reconciliation function confirmed the child remained alive beyond its configured deadline after SIGINT. The PR remains draft while cancellation cleanup and a regression test are added.

janicduplessis:
Both independent reviewers are clear at eabb87820bc67dd74ec4c7f4b131eab102df8c05 and verified the final committed files match their reviewed correction. The confirmed Ctrl-C finding is resolved; signal cancellation and immediate exit now stop only the verified SimSlim group, with bounded confirmation and retained claims when uncertain.

All local checks passed: format, lint, build, typecheck, knip, 4,248 unit tests (one existing skip), 84 e2e tests, and runtime. Real subprocess regressions verified SIGINT/SIGTERM/early-exit behavior; all fixture processes were cleaned. Native iOS discovery/boot/install/launch and the bounded opener were exercised separately, with exact owned-device cleanup. No SimSlim profile was applied.

PR #791, opened 2026-09-12
test: allow larger iOS runners for manual native QA

### LINKED ISSUE #790: Allow an opt-in higher-memory runner for native iOS QA
## Problem

Native iOS QA has no way to select a higher-memory hosted runner for a manual validation run. The standard macOS runner has 7 GB RAM; multiple simulators can coincide with sustained host pressure and boot timeouts.

## Evidence

The final [bare iOS cache job](https://github.com/appandflow/stim/actions/runs/34716978286/job/103615879932) failed its second simulator's 600-second cold-boot deadline, before the build race. Its timeline contains 42 warning-pressure samples and four unavailable readings on a 7 GB host. The same full cache suite passes on the local 16 GB Mac. These observations do not prove memory pressure is the sole cause.

[GitHub documents](https://docs.github.com/en/actions/reference/runners/larger-runners) an ARM64 `macos-latest-xlarge` runner with 14 GB RAM. Larger runners are billed, including for public repositories.

## Cause

The iOS job fixes `runs-on` to `macos-latest`, so a workflow dispatch cannot request more headroom without changing the default for every run.

## Fix idea

Add a manual, fixed-choice standard/large iOS runner input. Keep standard as the default for manual, scheduled, and PR runs; preserve the full suite matrix. Document the billing distinction. A paid validation run needs explicit authorization before dispatch.

janicduplessis:
## Description

Native iOS QA needs a way to test suspected host-capacity failures with more memory without moving every scheduled run onto a paid runner.

## Solution

Add a manual `ios_runner` choice. `standard` remains the default; only `large` selects GitHub's 14 GB ARM64 runner. Scheduled and PR runs retain the standard runner, and the full suite matrix is unchanged. Larger runners are [billed even for public repositories](https://docs.github.com/en/billing/reference/actions-runner-pricing); the QA guide makes that choice explicit.

## Test plan

Verify workflow syntax through GitHub's PR evaluation and run the repository checks. The paid runner has not been dispatched: after budget approval, run the combined native matrix with `ios_runner=large` and verify the actual runner, suite results, and memory artifacts.

Fixes #790.

### DIFF
--- a/.github/workflows/e2e-native.yml
+++ b/.github/workflows/e2e-native.yml
@@ -46,6 +46,11 @@ on:
         type: choice
         default: loop
         options: [loop, caches, pool, all]
+      ios_runner:
+        description: 'iOS runner: standard is free for public repos; large is billed'
+        type: choice
+        default: standard
+        options: [standard, large]
   pull_request:
     # `labeled` is what makes "add the label to run it" work; the others let a
     # push to an already-labeled PR re-run. The `if:` on each job is the actual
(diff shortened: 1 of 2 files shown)


janicduplessis:
Keeping the existing runners per the product decision. No paid runner was dispatched.

PR #853, opened 2026-09-15
fix: benchmark details scenario picker, terminal height, and timeline redesign

### LINKED ISSUE #852: Benchmark details page: scenario picker, terminal height, and status noise
## Problem

On `/benchmarks/details`:

- The terminal panel resizes every time a different command bar is selected because its height follows the output length (`min-height: 18rem; max-height: 32rem` on the `pre`), so the page jumps while clicking through commands.
- The Scenario picker offers `App readiness` and `Launch crash`, but readiness holds two distinct comparisons (JavaScript change and Native change) rendered as two side-by-side cards. Native change is not selectable on its own, and the run tabs mix variant and arm (`javascript / stim`, `native / control`).
- The environment row is labeled `Device` even though every published run uses an owned simulator or emulator.
- Several status fragments repeat protocol text that the methodology page already covers: `/ valid runs only`, the `Valid run` badge (only valid runs are ever rendered), the `Clock starts at...` timing paragraph, and the `"Agent / other" is time with no command active` note.
- The timeline section stacks three control rows over a bordered lane grid; adjacent bars run together and narrow bars show one-character fragments. The timeline and terminal use default browser scrollbars, which stand out against the panels.

## Evidence

Load `/benchmarks/details?benchmark=luna-rc12&run=javascript-stim`, select `stim guide lifecycle options` then `git worktree add`; the terminal section height changes with the output. The Scenario row has no Native change option.

## Cause

`website/src/pages/benchmarks/details.tsx` keys the scenario on `benchmark.suite` (`readiness` | `launch-crash`) rather than on the run variant, and `website/src/components/BenchmarkTimeline.module.css` sizes `.terminal pre` from content.

## Fix idea

Key the scenario picker on the run variant (`JS change`, `Native change`, `Launch crash`), show a single full-width comparison card for the selected scenario, fix the terminal section height with a viewport-based clamp, label the environment row `Simulator` / `Emulator` by platform, remove the repeated status text, collapse the timeline controls into one toolbar with scrubbing on the axis, replace lane borders with tick gridlines, give the terminal a thin themed scrollbar, and hide the timeline scrollbar (the timeline pans by wheel, drag, and keyboard, and its zoom control changes width).

janicduplessis:
## Description

The `/benchmarks/details` page had a few things working against it while clicking through a run:

- The terminal panel sized itself from its content (`min-height: 18rem; max-height: 32rem` on the `pre`), so the page jumped every time a different command was selected.
- The Scenario picker exposed `App readiness` / `Launch crash`, but readiness bundles two separate comparisons (JavaScript change, Native change) as side-by-side cards, and the run tabs mixed variant and arm (`javascript / stim`). There was no way to pick Native change directly.
- The timeline section stacked three control rows over a bordered lane grid, with adjacent bars running together and narrow bars showing one-character fragments.
- Several fragments repeated protocol text the overview page already covers: `/ valid runs only`, the `Valid run` badge (only valid runs are ever rendered), the `Clock starts at...` paragraph, and the `"Agent / other" is time...` note.

## Solution

The scenario picker is keyed on the run variant: `JS change`, `Native change`, `Launch crash`. `scenarioRun` in `benchmarkSelection.ts` picks the run for a scenario while keeping the current arm, so switching scenario keeps Stim vs Control (the URL moves between the readiness and launch-error benchmarks as needed), and switching model or platform keeps the scenario when the destination has it. The comparison section shows one full-width card for the selected scenario, and the run tabs list just that scenario's arms as `Stim` / `Control`.

The terminal section is a fixed-height flex column (`clamp(20rem, 65vh, 36rem)`) whose output `pre` fills the remainder and scrolls; expanding "Command context" eats into the output area instead of growing the box.

The timeline controls collapse into one toolbar, and the position slider becomes an invisible range input laid over the time axis so scrubbing happens on the axis itself (keyboard and `aria-valuetext` unchanged). Lane borders give way to tick gridlines and bars only render a label when wide enough to read it; the `aria-label` still carries the command.

The repeated protocol text listed above is removed, and the environment row reads `Simulator` or `Emulator` by platform instead of `Device`.

Also adds `.claude/launch.json` so the website dev server can be previewed from the Claude app.

## Test plan

- `pnpm test website/src` (57 tests) including new `scenarioRun` cases for arm preservation and fallback.
- Load `/benchmarks/details?benchmark=luna-rc12&run=javascript-stim`, click through commands with different output lengths: the terminal section stays at the same height.
- Switch scenario `JS change` -> `Launch crash` -> `Native change`: the arm stays on Stim and the single card and run tabs follow.
- Switch platform to Android: the environment row reads `Emulator`.
- Drag on the time axis: the playhead follows and the terminal shows the command under the cursor. Tab to the axis and use arrow keys.
- Checked at 400px width (toolbar wraps to two rows) and in the dark theme.

Pickers, card, and environment:

Timeline and terminal:

Fixes #852


janicduplessis:
Fresh review (Claude + Codex) findings addressed in cefd2fbe8 and 8e2cd6b: dropped the unused `defaultRun` preferred-id parameter and its test, aligned the last axis tick after the scrubber became the axis's last child, snapped `-`/`+` zoom to half steps so a pinch value such as 1.53x no longer yields 2.0300000000000002x, and capped the expanded command context at 45% of the terminal panel so output stays visible. Kept as decided: the timeline scrollbar is hidden (wheel, drag, keyboard, and the zoom control still pan it) and `.claude/launch.json` is included. Not addressed: `.eventDetail` (agent note / marker panel) keeps its own smaller height, so switching between a note and a command still moves the page.

PR #874, opened 2026-09-19
chore: raise the Node floor to 22.12.0 and take commander 15 and chalk 6

### LINKED ISSUE #867: Node 20 is end of life: raise the published runtime floor to Node 22.12.0 and take commander 15 and chalk 6
## Problem

Node 20 reached end of life, but the five published packages still advertise it
as a supported runtime and CI still proves it:

- every `package.json` declares `"node": "^20.19.4 || >=22.12.0"`
- `test/runtime-floor.mjs` asserts that exact range and is executed under Node
  20.19.4 by the `published runtime (node 20.19.4)` job in `.github/workflows/ci.yml`
- `packages/stim-cli/README.md`, `website/docs/requirements.md`, `docs/e2e-and-ci.md`,
  `AGENTS.md`, `RELEASE.md` and the `STIM_NODE_UNSUPPORTED` guide text all repeat
  the Node 20 floor
- `packages/stim-cli/tsdown.config.ts` emits for `node20.19`

Advertising an end-of-life runtime means the floor job proves something nobody
should run, and it blocks dependency majors: `commander@15` requires Node
`>=22.12.0` and `chalk@6` requires Node `>=22`, so neither can be adopted while
Node 20 is in the range. The CLI is pinned at `commander@^13.1.0` and
`chalk@^5.4.1`, two and one majors behind.

## Evidence

```
$ grep -rn '"node": "\^20' packages/*/package.json
packages/expo-build-cache/package.json:34:    "node": "^20.19.4 || >=22.12.0"
packages/core/package.json:43:    "node": "^20.19.4 || >=22.12.0"
packages/cache/package.json:30:    "node": "^20.19.4 || >=22.12.0"
packages/stim-cli/package.json:62:    "node": "^20.19.4 || >=22.12.0"
packages/metro/package.json:38:    "node": "^20.19.4 || >=22.12.0"

$ npm view commander@15.0.0 engines
{ node: '>=22.12.0' }
$ npm view chalk@6.0.0 engines
{ node: '>=22' }
```

Node 20 entered end of life in April 2026 (nodejs/release schedule). Commander 15
release notes: "Commander 15 requires Node.js v22.12.0 or higher (for
`require(esm)`)". Chalk 6 release notes: "Breaking: Require Node.js 22".

## Cause

The floor was chosen when Node 20 was still in maintenance. `20.19.4` and
`22.12.0` are both the first release on their line that ships unflagged
`require(esm)`, which is what `test/runtime-floor.mjs` proves when it calls
`require('@stim-cli/core')` against an ESM-only build. Nothing else ties the
packages to Node 20.

## Fix idea

Drop the Node 20 arm and keep the Node 22 arm exactly where it is: engines
`>=22.12.0`, runtime-floor pinned at 22.12.0, the CI job renamed to match, and
every document that states the floor updated in the same commit. 22.12.0 stays
the right number for the same reason it was already the Node 22 arm -- it is the
lowest Node 22 release where `require(esm)` works, which the floor test depends
on, and it is also commander 15's own floor.

With Node 20 gone, take `commander@^15` and `chalk@^6` and adapt the call sites
for their breaking changes. No other dependency bumps.

janicduplessis:
## Description

Node 20 is end of life, but the five published packages still declared
`^20.19.4 || >=22.12.0`, the CI runtime-floor job ran `test/runtime-floor.mjs`
under Node 20.19.4, and six documents repeated the two-arm sentence. Proving a
runtime nobody should be on is the visible cost; the invisible one is that it
pinned two dependencies -- `commander@15` declares `>=22.12.0` and `chalk@6`
declares `>=22`, so the CLI was stuck on `commander@13` and `chalk@5`.

**On the diff size:** only 4 of the ~940 changed lockfile lines are this PR's
(two dependency specifiers, two package entries). The rest is pre-existing pnpm
peer-suffix churn in the website's webpack tree. It is not caused by these bumps
-- on clean `origin/main` with pnpm 12.0.0, bumping an unrelated range (`ws`
`^8.21.3` -> `^8.21.2`, same resolved version) rewrites the same 771 lines.

Fixes #867.

## Solution

`>=22.12.0` is not a new number: it is the existing Node 22 arm, kept. That arm
was already 22.12.0 because it is the lowest Node 22 release with unflagged
`require(esm)`, which is what `test/runtime-floor.mjs` depends on. So this drops
an arm rather than moving a floor -- engines, the runtime-floor assertion, the CI
job's name and `node-version`, the tsdown `target` in all five packages
(`node20.19` -> `node22.12`),
and every document that states the floor (`AGENTS.md`, `RELEASE.md`,
`docs/e2e-and-ci.md`, `packages/stim-cli/README.md`,
`website/docs/requirements.md`, and the `environment` guide topic's "Unsupported
engine" remedy). Historical `docs/releases/*.md` are left alone; they describe
what shipped. `website/package.json` keeps its scaffolded `"node": ">=20.0"`:
it is a dev-only constraint on the docs site, not a published floor, and the
root's `>=22.18.0` already binds tighter.

commander 15 and chalk 6 needed no call-site changes. commander 15's `--no-*`
break (tj/commander.js#2405) does not reach any flag here, which the parity check
below confirms rather than assumes. chalk 6's breaks touch nothing: the CLI never
calls `ansi256`, and its only `FORCE_COLOR` use sets `'0'` in child-process env,
which still means "no color".

The one user-visible difference is commander's excess-argument error, which now
names the extras. Nothing asserts on or parses this string:

```
error: too many arguments for 'stats'. Expected 0 arguments but got 2: extra1, extra2.
```

## Test plan

- **`node test/runtime-floor.mjs` under a real Node 22.12.0** (downloaded
  `node-v22.12.0-darwin-arm64`, not the machine's 22.22.1): passes. This is the
  check that matters -- it loads every published ESM entry point through
  `require()` and runs `dist/cli.mjs --version` and `--help`, so it also proves
  commander 15 works on the new floor and not merely on a newer Node.
- **commander 13 vs 15 parity** over the real `ios` flag set (`--no-metro-check`,
  `--no-build-cache`, `--device [udid]`, `--wait <seconds>`, `--no-wait`):
  identical `opts()` for `[]`, `--device`, `--wait 30`, `--no-wait`, each lone
  negated flag, and `--wait 30 --no-wait`.
- **Built CLI error paths** against `dist/cli.mjs`: unknown option, excess
  arguments, and an `InvalidArgumentError` from the `--remote` parser all still
  exit 1 with the expected text.

Full required set green locally on the pushed commit: `format:check`, `lint`,
`build`, `typecheck`, `knip`, `test` (146 files, 4501 passed, 1 skipped),
`test:runtime`, `test:e2e` (95/95). `test:compat` was not run -- this change
makes no native tool calls.

The unit suite flakes on this machine under full 146-worker concurrency
(`engine-simslim.test.ts` and `sim-pool.test.ts`, child-process spawn timeouts);
those two files pass 39/39 in isolation and the flake reproduces on `origin/main`,
so it is not from this change.


janicduplessis:
Reviewed by a fresh agent that did not implement the change. One substantive finding, now fixed in `80c849c7e`: the first revision moved only `packages/stim-cli/tsdown.config.ts` to `target: 'node22.12'` and left `core`, `cache`, `metro` and `expo-build-cache` at `node20.19` -- their configs are `tsdown.config.mts`, which the original sweep's glob missed. All five now match the engines floor.

Everything else checked out. The reviewer independently confirmed the commander 13-vs-15 `--no-*` parity, that no chalk 6 break reaches a call site, that `target: 'node22.12'` is actually honored by rolldown rather than silently ignored (a downlevel-size comparison against a bogus target), that no stale Node 20 reference remains outside the historical `docs/releases/*.md`, and that renaming the `published runtime` job breaks no required-status-check gate (`main` has no branch protection and its ruleset declares no `required_status_checks`).

PR #887, opened 2026-09-19
fix: make the Windows Android e2e loop get past stim start

### LINKED ISSUE #881: The Windows Android e2e loop stops at stim start: the Metro readiness probe requires lsof
## Problem

The non-blocking `windows android e2e` job cannot complete the Android loop.
`stim start` launches Metro, Metro's own records say `server_listening` and
`initialize_done`, and then `start` refuses with

```
STIM_METRO_TIMEOUT: The dev server did not answer on port 8082 within 240s.
```

Because `stim start` never returns, `stim android` never runs, no Gradle build
happens, and the job's `Assert Stim produced an APK` step -- the one step that
decides the job's colour -- fails with "stim android produced no APK under the
e2e temp root".

The goal is the whole loop on `windows-latest`: fixture created, `stim start`
ready, `stim android` building with Gradle, installing on the WHPX emulator,
launching, and the APK assertion green.

## Evidence

- Run https://github.com/appandflow/stim/actions/runs/35420216961, job
  `windows android e2e (non-blocking)`, step `Android loop e2e (bare)`. The
  `metro: server_listening` and `bare-inproc dev server listening on port 8082`
  records sit directly above the refusal.
- Reproduced identically on two separate Windows runs.
- `emulator -accel-check` on the same runner reports
  `WHPX(10.0.26100) is installed and usable`, so the emulator half is not the
  blocker.

## Cause

Three separate Windows gaps, in the order the loop hits them.

**1. The readiness probe requires `lsof`.** `waitForMetro`
(`packages/stim-cli/src/commands/start.ts:947`) polls `resolveProjectMetro`,
whose first act is
`getExecutor().runQuiet('lsof -nP -iTCP:<port> -sTCP:LISTEN -t')`
(`packages/stim-cli/src/metro.ts:83`). `lsof` does not exist on Windows, so
`runQuiet` returns `null`, the listener list is empty, and the function returns
`{ missing: true }` forever -- before it ever probes `/status`. The same call
gates `stim android` (`commands/android.ts:806`) and `stim status`.

The cwd half has no Windows answer either: `processCwd`
(`packages/stim-cli/src/metro.ts:48`) reads `/proc/<pid>/cwd` on Linux and
shells out to `lsof` everywhere else, and no cheap Windows tool reports another
process's working directory.

This supersedes the suspicion recorded in #871, that
`supervisor/server-bare.ts` binds `localhost` while `ports.ts` probes
`localhost` and the two pick different loopback families. Metro's `runServer`
is called with no `host`, so it binds the unspecified address and is reachable
on both `127.0.0.1` and `::1`; the probe never gets far enough to matter.

**2. Android SDK tools are not named for Windows.** `SDK_TOOL_LOCATIONS`
(`packages/stim-cli/src/devices/android.ts:79`) joins the bare names `emulator`,
`adb` and `avdmanager`. On Windows those files are `emulator.exe`, `adb.exe`
and `avdmanager.bat`, so `existsSync` fails and `androidToolPath` falls back to
the bare name and a PATH lookup. `cmdline-tools\latest\bin` is not on PATH, so
AVD creation cannot find `avdmanager` at all.

**3. The exec wrapper cannot spawn a `.bat`/`.cmd` shim.**
`packages/stim-cli/src/exec.ts:62` spawns without `shell`, and Node has refused
to spawn `.bat`/`.cmd` that way since 18.20.2 (the CVE-2024-27980 fix). Two
spawns on the Android path are batch wrappers on Windows: `gradlew.bat`
(`engine/gradle.ts:42` via `gradlewPath`) and `avdmanager.bat`
(`devices/android.ts:233`, `createAvd`).

Related but not on the e2e path: the nine
`runQuiet('command -v <tool>')` PATH lookups are a POSIX shell builtin and
always report "not installed" under `cmd.exe`.

## Fix idea

1. Let `resolveProjectMetro` recognize this workspace's own supervisor from its
   `ProcessRecord` -- `supervisor.port === port` plus
   `inspectProcessIdentity(supervisor) === 'same'` plus a `/status` answer --
   before it asks the OS which pid holds the port. That record is the
   authoritative ownership signal the "Locked state" rule already names; the
   pid-and-cwd route stays for a Metro this workspace did not start.
2. Give `androidToolPath` the Windows extension per tool.
3. Make the single exec wrapper resolve `.cmd`/`.bat` shims.

All three must leave macOS and Linux byte-identical.

janicduplessis:
## Description

On `windows-latest` the Android e2e loop stopped at `stim start`, which
refused with `STIM_METRO_TIMEOUT` while Metro's own records said
`server_listening`. The readiness probe runs `lsof`, which does not exist on
Windows, so it answered `missing` until it timed out. The same probe gates
`stim android` and `stim status`, so nothing downstream ran: no Gradle build,
no APK, and the job's `Assert Stim produced an APK` step -- the one step
without `continue-on-error` -- failed.

Six more Windows gaps sat behind it, each found by running the loop again
once the one above it was fixed. The last three were found on an interactive
`windows-latest` host rather than through CI.

Fixes #881.

## Solution

- **The readiness probe.** Listener discovery moved behind `listeningPids`,
  which asks lsof everywhere and falls back to `netstat -ano` when win32 lsof
  produces nothing. A listening TCP row is recognized by its foreign address
  carrying port 0 rather than by the `LISTENING` state column, which Windows
  localizes.
- **Android SDK tool names.** Windows installs `emulator.exe`, `adb.exe` and
  `avdmanager.bat`; `SDK_TOOL_LOCATIONS` joined the bare names, so
  `existsSync` missed all three and `androidToolPath` fell back to a PATH
  lookup. `cmdline-tools\latest\bin` is not on PATH, so AVD creation could
  not find `avdmanager` at all. Only the on-disk path carries the extension;
  the bare-name fallback stays bare, because Windows appends `.exe` itself
  and `cross-spawn` resolves a `.bat` through PATHEXT.
- **The exec wrapper.** `command -v` is a POSIX shell builtin, so the nine
  lookups behind it reported "not installed" for ccache, agent-device,
  simslim, the EAS CLI and the tunnel providers under `cmd.exe`; and Node has
  refused to spawn `.bat`/`.cmd` without a shell since 18.20.2 (the
  CVE-2024-27980 fix), which both `gradlew.bat` and `avdmanager.bat` need.
- **The Android boot budget.** WHPX is the only Android accelerator on a
  Windows host, and the emulator was still running its boot animation when
  the 120s fresh-AVD wait expired. Both Android waits take the 600s the iOS
  boot wait already uses, on win32 only. The AVD then took 9m38s to create
  and boot.
- **CMake object paths over 260 characters.** See below.
- **`stim stop`.** It signalled the negative leader pid, the POSIX way to
  reach a process group. Windows has no equivalent, so `process.kill` threw
  and `stop` refused, leaving the port reservation and the device behind.
- **The e2e harness.** It identified processes with
  `ps -ax -o pid=,lstart=,command=`. The Git-for-Windows `ps` on PATH
  answered `unknown option -- x`, so the run died at the first
  `recordWorkspace`. `Win32_Process` carries the same three fields.

### The 260-character path limit

The Android NDK pins CMake 3.22.1, whose ninja is not long-path aware, and
CMake names an object for a source outside the CMake source directory after
that source's mangled absolute path. React Native's codegen builds exactly
those, so `react_codegen_safeareacontext` produced a 415-character path and
`ninja: error: ... Filename longer than 260 characters`.

Measured on a `windows-latest` host with the fixture at `D:\f`, a plain
Gradle build's longest object path is 245 of the 259 available -- and the
`stim-<profile>` staging directory the Gradle shim inserts costs 22 of that
headroom, which is what turns a tight fit into a failure. The shim now pays
for the room it takes: on Windows it sets `CMAKE_OBJECT_PATH_MAX`, which
makes CMake replace the mangled source directory with a hash. The same build
then produced 238 characters. The e2e fixture root feeds every one of those
paths, so the driver also gives its work directory a short name on Windows
and the job puts it at `D:\e`.

### The ownership half of the readiness probe

Windows has no cheap way to read another process's working directory, so the
cwd check in `resolveProjectMetro` needed a second answer rather than a
second tool. An unreadable cwd is no longer an automatic refusal: the
listener is accepted when it is this workspace's own recorded supervisor --
or the `serverPid` it recorded for its Expo child -- and that record's
process token still proves the holder live.

**This is strictly narrower than the cwd check it stands in for.** A listener
this workspace never recorded is still refused, so foreign-Metro detection
and port stealing are unchanged, and on macOS and Linux the cwd is readable
so the branch is unreachable.

Two consequences worth stating. On Windows `processCwd` is always null, so a
Metro this workspace did not start is permanently `notOurs`: Stim will move
its reservation to another port rather than adopt a dev server it cannot
identify. That is the fail-closed direction and it is intended, but it does
mean the "a dev server already answers, started outside Stim" path never
fires there. And the slot log file rename is forward-only: a workspace that
already holds colon-named slot logs keeps them, and because the logs query
enumerates every `*.ndjson`, those records stay in `stim logs` output until
the workspace's logs directory is cleared.

### What still fails, and why it needs a decision

The job is green and its `Assert Stim produced an APK` step passes, but the
non-blocking `Android loop e2e (bare)` step still exits 1, always after the
part this PR set out to fix has passed. Where it stops varies between runs.
One point is deterministic and one is a race on a loaded Windows runner: a
run stopped at `adb install ... cmd: Can't find service: package` 941ms after
the boot wait cleared, so the boot props are set before the package manager
is up. The deterministic one is the named-slots phase.

Reproduced on the interactive host: when a slot repeats on a device it
already has, `device.systemImage` is missing from the record, so
`resolveAndroidRun` gets no ABI, the fingerprint key changes (`hit` becomes
`miss` for the identical fingerprint), and Gradle builds every ABI instead of
the emulator's. The `arm64-v8a` codegen object path then exceeds 260
characters again -- and this time `CMAKE_OBJECT_PATH_MAX` cannot save it:
CMake hashes the mangled source directory only when the hashed path fits
under the maximum, and for
`react_codegen_safeareacontext/.../ComponentDescriptors.cpp.o` the hashed
form is still 304 characters at a 32-character project root. Lowering the
maximum to 200 changes nothing, which is the measurement that says this is a
ceiling rather than a tuning problem.

Two decisions sit behind it, neither of which belongs in this PR: the
ABI-less build is #892, which is platform-neutral and on macOS only costs a
redundant all-ABI compile, and what Stim does when a project path leaves no
room for the NDK's non-long-path-aware ninja is #893.

### Not fixed here

The release APK-swap path is still POSIX-only on Windows, for reasons wider
than this change: `engine/apk-swap.ts` shells out to `cp`, `mv` and `zip`,
and signs with `apksigner.bat` through `runFile`, which still uses
`execFileSync` and so cannot launch a batch wrapper. The debug Android loop
this job drives does not touch it. `named-ports.ts` also still calls `lsof`
directly for `stim ports`. Both are #894.

Two deliberate choices. `findExecutable` went on the `Executor` interface
rather than staying a free function so the tests that used to stub
`runQuiet('command -v ...')` stub it instead; a mock that forgets it now
fails loudly instead of silently reporting the tool absent. And `spawn` goes
through `cross-spawn` rather than a local quoting helper, because
hand-rolling cmd.exe escaping is what CVE-2024-27980 was about.

## Test plan

- The loop was run end to end four times on an interactive `windows-latest`
  host: fixture created, two warm worktrees, `stim start` ready in
  `bare-inproc` mode, `stim android` compiling with Gradle in 2-3.5m,
  `app-debug.apk` installed on the WHPX emulator, launched, and
  `verify bundle loaded, process alive` -- then the second worktree
  installing from the build cache with no compiler invocation in its log,
  and `stim stop` shutting the supervisor, collector and emulator down.
- `windows android e2e (non-blocking)` is green on this branch, every step
  including `Assert Stim produced an APK`
  (`stim android produced ...\app-debug.apk (36.5 MB)`).
- `signalProcessTree` was exercised against the real `taskkill` on that host:
  a parent with a grandchild reports `before parent=true child=true`,
  `signal=true`, `after parent=false child=false` -- `/T` reaches the tree
  that `process.kill(pid, 'SIGTERM')` would have orphaned -- and a pid that
  does not exist returns false.
- macOS control, locally: `pnpm run format:check`, `pnpm run lint`,
  `pnpm run build`, `pnpm run typecheck`, `pnpm run knip` clean; `pnpm test`
  147 files, 4520 passed, 4 skipped.
- Unit coverage for the probe: `parseNetstatPids` picking the listening row
  for a port out of real `netstat -ano` output, `listeningPids` falling back
  to netstat only on win32, and `resolveProjectMetro` accepting an
  unreadable-cwd listener that is the recorded supervisor while still
  refusing one that is not.


janicduplessis:
The localhost/IPv6 binding named as the next thing to fix in #871 is not the cause. Metro's `runServer` is called with no `host`, so it binds the unspecified address and answers on both `127.0.0.1` and `::1`; `stim start` never reached the `/status` probe because `resolveProjectMetro` asks `lsof` for the listening pid first and returns `missing` when that produces nothing.

Issue #901, opened 2026-09-19
Native e2e: smoke on every push to main, opt-in labels per suite on pull requests


janicduplessis:
## Decision

Pull-request feedback stays fast (the unit lanes, about 5 minutes). Native e2e runs:

- on every push to `main`: the **smoke** suite (`run-native-e2e.mjs --smoke`, from #899) on iOS (`xcode-27`, Expo), Linux Android (Expo) and Windows Android (bare);
- nightly: the loop suite, as today;
- on `workflow_dispatch`: the `suite` input, `smoke` added;
- on a pull request: only what its labels ask for. Labels exist in the repo: `e2e-smoke`, `e2e-loop`, `e2e-caches`, `e2e-pool`, `e2e-all`. Several labels union. Adding a label re-triggers via the `labeled` event. (The old `e2e-native` label was never created, so that gate never fired.)

## Shape

- The Windows Android job moves from `ci.yml` into `e2e-native.yml`, so every native gate lives in one workflow; `ci.yml` keeps the fast lanes including `test (windows)`. The windows-branch-name heuristic goes away in favour of labels.
- A `plan` job computes the suite list per platform from event, labels and input through a checked-in `scripts/e2e-plan.mjs` with a unit test (`scripts/e2e-plan.test.mjs`), and the three platform jobs read `needs.plan.outputs.<platform>` as their `suite` matrix, skipping when the list is empty. Per platform: iOS smoke/loop/caches/pool (Expo only), Linux Android smoke/loop/caches, Windows Android smoke/loop. Smoke is one framework per platform.
- The smoke restores the cross-run build cache the loop already persists so it is warm when the native fingerprint is unchanged; the Windows lane gets the same cache.
- `docs/e2e-and-ci.md` and the workflow headers describe the gate and the labels.

## Done when

- A push to main runs three smoke jobs and they pass.
- A pull request with no label runs no native job; with `e2e-smoke` it runs the three smoke jobs; with `e2e-loop` the loop matrix as the nightly runs it.
- The nightly and dispatch behave as before.

janicduplessis:
Claimed: @janic/native-smoke-pr (based on #899's branch until it merges)

janicduplessis:
Claimed: @janic/e2e-labels-plan

PR #902, opened 2026-09-19
ci: run the native smoke suite on every pull request on macOS and Linux

janicduplessis:
Closes #901. Stacked on #899 (base branch set to it; will retarget main once #899 merges).

Pull requests never built or launched anything natively on macOS or Linux: `e2e-native.yml` ran nightly, on dispatch, or on a pull request carrying the `e2e-native` label, which none has carried. This adds a **smoke** suite to that workflow, driven by `run-native-e2e.mjs --smoke` from #899: one worktree, `stim start`, build, install, launch, readiness, `stim stop`, cleanup checks. Expo framework only (bare is already excluded on iOS 27), so one `ios expo (smoke)` job on `xcode-27` and one `android expo (smoke)` job on Ubuntu with KVM, blocking.

**Gate.** The job-level `if:` gates are replaced by the matrix `suite` expression: an unlabeled pull request selects `smoke`, a labeled one `loop`, the nightly `loop`, and dispatch its input (`smoke` is a new option). Pull requests touching only `docs/`, `website/` or Markdown skip the workflow via `paths-ignore`.

**Speed.** The smoke restores the same cross-run build cache the loop already persists with `actions/cache`, so a pull request that leaves the native fingerprint alone installs from cache (about 8 minutes iOS, 5 Linux); one that changes it pays the cold build. The Windows Android lane in `ci.yml` gets the same cache (`D:\e\build-cache`, keyed like the others) and the same warm-cache detection, bringing its typical run from ~25 to ~14 minutes, the emulator boot being the floor.

`docs/e2e-and-ci.md` and both workflow headers describe the new gate.


janicduplessis:
Superseded: the gate is changing to push-to-main plus opt-in labels (e2e-smoke, e2e-loop, e2e-caches, e2e-pool, e2e-all) with the Windows job moving into e2e-native.yml; a fresh PR against #901 follows.

PR #921, opened 2026-09-20
fix: Android repeated slot keeps its ABI on Windows, boot wait covers the package manager

### LINKED ISSUE #892: Repeated named slot loses device.systemImage, so Android builds every ABI and the fingerprint misses
## Problem

When a named slot repeats on a device it already owns, `device.systemImage` is missing from the slot's device record, so `resolveAndroidRun` gets no ABI. Consequences:

- Gradle builds every ABI instead of the emulator's (`arm64-v8a` and friends), a redundant compile on every platform.
- The fingerprint key changes with the ABI set, so the identical fingerprint reports `miss` instead of `hit`.
- On Windows the extra ABIs push codegen object paths past the 260-character limit (see the path-length issue), which is how it surfaced.

Reproduced on the interactive `windows-latest` host during #887 in the driver's named-slots phase; the same record shape is produced on macOS, where it only costs the extra compile.

## Done when

- A repeated slot's device record keeps `systemImage` (or the ABI is re-derived from the running emulator), and `resolveAndroidRun` passes a single ABI in both the first and the repeated slot run.
- A unit test covers the repeated-slot record, and the named-slots phase of `test/e2e/native/run-native-e2e.mjs --platform android` reports `hit` on the second run.

Origin: #887 "What still fails, and why it needs a decision".

### LINKED ISSUE #897: Android boot wait can return before the package manager is up, so the first adb install fails
## Problem

On the final #887 run the non-blocking loop step stopped at worktree 2's `adb install` with `cmd: Can't find service: package`, 941 ms after the boot wait cleared. The emulator sets its boot properties before the package manager service is registered, so `waitForBoot` can return while `pm` is not yet reachable, and the first `adb install` fails. Seen on a loaded `windows-latest` runner; the mechanism is platform-neutral (Android boot order), so a slow macOS or Linux host can hit it too. Pre-existing; #887 touched neither install nor boot detection.

## Fix idea

After the boot property check, also wait until the package manager answers, for example `adb shell cmd package list packages` (or `pm path android`) returning 0, with the same budget as the boot wait. Keep the boot-property check as the first gate so a fast host pays one extra ~100 ms probe.

## Done when

- `waitForBoot` (engine/device.ts) does not return until `pm` answers, with a unit test that a device reporting `sys.boot_completed=1` but a failing `pm` probe keeps waiting.
- The `android e2e (windows)` loop step no longer stops at `adb install` across three consecutive runs.

Origin: #887 final run, noted in its PR body.

janicduplessis:
## Description

Two Android fixes that surfaced in the Windows e2e loop (#887) but are platform-neutral.

- #892: **an AVD created on Windows stores `image.sysdir.1` with backslashes, and `parseAvdSystemImage` only split on `/`.** A repeated named slot re-derives `device.systemImage` from the running AVD's `config.ini`, so on Windows it got `system-images\android-36\google_apis\x86_64\` back verbatim: no ABI, Gradle built every ABI, and the cache key (which includes `abi`) changed for an identical fingerprint, so the second run reported `miss`.
  The same parser feeds park-and-adopt matching, so parked AVDs on Windows never matched either.

- #897: **`waitForBoot` returned on `sys.boot_completed=1`, but the emulator sets that before the package manager service is registered.** The first `adb install` then failed with `cmd: Can't find service: package` (941 ms after the boot wait cleared on a loaded runner).

## Solution

- `parseAvdSystemImage` accepts either separator. The stored device record is unchanged; the ABI is still re-derived from the AVD each run, now correctly on Windows too.

- After the boot-property gate, `waitForBoot` keeps polling until `adb -s <serial> shell pm path android` answers with a `package:` line, inside the same boot budget and per-probe timeout. A device whose package manager is already up pays one extra probe. The timeout diagnostic gains a `packageManager` field so a boot that stalls after the properties is distinguishable from one that never set them.

## Test plan

- Unit (`engine-device.test.ts`): create an owned AVD for slot `second`, rewrite its `config.ini` with backslashes, repeat the slot on the running emulator. `systemImage` is preserved, `androidBuildOptions` yields the same single ABI, and `buildCacheKey` is identical between runs. Failed before the parser change with `Received: "system-images\android-36\google_apis\arm64-v8a\"`.
- Unit (`sim-android.test.ts`): boot completed but `pm` fails twice before answering keeps waiting, then `ok`; `pm` already up costs exactly one probe; `pm` never answering times out with `packageManager: ''`.
- Real tool (invariant 9): on a headless API 31 AVD, `adb shell pm path android` prints `package:/system/framework/framework-res.apk` (exit 0) once up and `adb: device offline` (non-zero) before.
- `node test/e2e/native/run-native-e2e.mjs --framework bare --platform android` on macOS: named-slots result in the comment below. The `e2e-loop` label runs the same loop on Linux and Windows.

Fixes #892
Fixes #897


janicduplessis:
Local loop on macOS (arm64, API 36 images), `node test/e2e/native/run-native-e2e.mjs --framework bare --platform android` at ab8f976: `PASS bare-android`, all five cleanup checks clean. Named-slots phase:

```
device      stim-e2e-1-3b706606bc9f-second created (36.3s)
fingerprint 374149.. hit (1.3s)
install     app-debug.apk -> emulator-5558 (2.1s)
fingerprint 374149.. hit (851ms)              <- repeated slot `second`
device      stim-e2e-1-3b706606bc9f-second (emulator-5558) booted (101ms)
install     unchanged (emulator-5558 already has this build) (95ms)
device      stim-e2e-1-3b706606bc9f-third created (35s)
fingerprint 374149.. hit (1.4s)
device      shut down stim-e2e-1-3b706606bc9f-second
device      stim-e2e-1-3b706606bc9f-second prepared (3.3s)
fingerprint 374149.. hit (873ms)              <- resumed after `stop --slot second`
SLOT PROOF: three assignments, reuse, cache hits, attributed logs and isolated stop.
```

Cache key on every run is `3741497f...-debug-sim-arm64-v8a`, so the repeated slot keeps the single ABI. The Windows evidence for #892 is the `android bare (loop, windows)` job from the `e2e-loop` label.

janicduplessis:
`e2e-loop` run at ab8f976 (https://github.com/appandflow/stim/actions/runs/35486455834): `android bare (loop)`, `android expo (loop)`, `ios expo (loop)` and `android bare (loop, windows)` all pass.

Windows named-slots phase (the one #892 broke), every run `hit` on the same key, no `Can't find service: package` on any of the six installs:

```
device      stim-e2e-1-97b836005714-second created (12m51s)
fingerprint bfead0.. hit (12.8s)
install     app-debug.apk -> emulator-5558 (24.1s)
fingerprint bfead0.. hit (9s)                  <- repeated slot `second`
install     unchanged (emulator-5558 already has this build) (2s)
fingerprint bfead0.. hit (15.8s)               <- slot `third`
device      stim-e2e-1-97b836005714-second prepared (23.1s)
fingerprint bfead0.. hit (26.3s)               <- resumed after `stop --slot second`
SLOT PROOF: three assignments, reuse, cache hits, attributed logs and isolated stop.
PASS bare-android
```

Linux named-slots: `374149.. hit` on all four slot runs, `SLOT PROOF`, `PASS bare-android`.

PR #922, opened 2026-09-20
fix: release APK swap and named ports without cp, mv, zip, lsof on Windows

### LINKED ISSUE #894: Windows: release APK swap and stim ports still depend on cp/mv/zip and lsof
Windows gaps left out of #887 because the debug Android loop does not touch them:

- `engine/apk-swap.ts` (release APK swap) shells out to `cp`, `mv` and `zip`, and signs with `apksigner.bat` through `runFile`, which still uses `execFileSync` and cannot launch a batch wrapper. Needs the same treatment as the debug path: `fs` operations instead of coreutils, a zip implementation that exists on Windows (or the JDK's `jar`), and `cross-spawn` for `.bat`.
- `named-ports.ts` still calls `lsof` directly for `stim ports`; `metro.ts` already has the `netstat -ano` fallback (`listeningPids`) to reuse.

Done when `stim android --release` (APK swap path) and `stim ports` run on the `windows-latest` debug host (#888 workflow) and the corresponding unit tests have win32 coverage.

Origin: #887 "Not fixed here".

janicduplessis:
## Description

Two Windows gaps from #894, both outside the debug Android loop that #887 fixed:

- The release APK swap (`stim android --variant release` on a cache hit) shelled out to `cp -c`, `mv`, `zip -d` and `zip -0 -r`. None of those exist on a stock Windows host, so the swap failed at the copy step and the command fell through to a full Gradle build.
- `named-ports.ts` inspected listeners with `lsof` and described them with `ps -p`. `devices/reclaim.ts` calls `clearNamedPorts(path, { stop: true })`, so `stim worktree remove` (and reclaim through `gc`) threw `spawn lsof ENOENT` on Windows whenever the workspace held a named port, not just `stim ports stop`.

## Solution

The copy keeps `cp -c` on darwin (libuv implements `COPYFILE_FICLONE` only on Linux, so `copyFileSync` reads every byte there: 2.1 s vs 3 ms for a 1.5 GB file on APFS) and falls back to `copyFileSync` when `cp` refuses; Linux and Windows use `copyFileSync` directly. The Hermes output is `renameSync`d into place.

The archive update uses the JDK's `jar` tool, resolved from `JAVA_HOME/bin` when set and PATH otherwise. A JDK is always present because Gradle needs one, and no pure-JS zip is in the dependency tree. `jar --update --file <apk> --no-compress -C <stage> assets` replaces `assets/index.android.bundle` in place (so the separate `zip -d` step and its "nothing to do" tolerance are gone) and stores the new entry. Verified on a synthetic APK that the entry set and compression methods match the old zip output (`resources.arsc` stays stored, an existing `META-INF/MANIFEST.MF` is left alone); only entry order differs, which `zipalign` and `apksigner` ignore.

Named ports go through `listeningPids()` (netstat) only on win32. POSIX keeps the previous `runFile('lsof', ..., { timeoutMs: 5000 })` where exit 1 with no output means no listeners and anything else keeps the allocation, because `listeningPids` maps every lsof failure to "no listeners" and `ports stop` would then release under a live server. A missing lsof is refused once per `clearNamedPorts` with `Cannot stop named ports: lsof is not installed.` before any label is touched. The listener description is `ps -p` on POSIX and `tasklist /FI "PID eq N" /FO CSV /NH` (image name) on win32; the signal goes through `signalProcessTree` (`process.kill` on POSIX, `taskkill /T /F` on win32).

## Test plan

- Unit: `engine-apk-swap.test.ts` covers the new tool order, the exact `jar` argv, `jarPath` resolution, `cp -c` and its fallback on darwin, and the no-`cp` path on linux/win32. `named-ports.test.ts` mocks the executor on `darwin` (lsof via `runFile` with the timeout, ps, `process.kill`) and `win32` (netstat, tasklist, taskkill) and asserts the argv and log lines for dry run and stop, plus the missing-lsof and lsof-failure refusals.
- Real tool: `jar --update --file x.apk --no-compress -C stage assets` against a synthetic APK on macOS (JDK 17), compared with `unzip -v` against the old `zip -d` + `zip -0 -r` result: `resources.arsc` and `lib/*.so` stored, `classes.dex` deflated, same `assets/` directory entry, manifest untouched.
- Windows runner (windows-debug.yml, this branch built on the host): `stim android --variant release` twice (miss, then the cache hit that takes the APK swap path), `stim ports get`, and `stim worktree remove` while the workspace holds a named port. Evidence in a comment below.

Fixes #894


janicduplessis:
Windows evidence (windows-debug.yml run 35485763625, `windows-latest`, this branch at 6efe611 built on the host, Temurin JDK 17 with `JAVA_HOME` set, build-tools 36, bare RN fixture from `run-native-e2e.mjs --fixture-only`).

**Release APK swap.** `stim android --variant release` in a warm worktree: first run was a cache miss (Gradle 7m36s, launch verified). Appended a line to `App.tsx` and ran it again:

```
  fingerprint bfead0.. hit (1.7s)
  swap        regenerating this workspace's JS for the cached release APK
  swap        hermes bytecode repacked (store), zipaligned and re-signed (25.4s)
  install     app-release.apk -> emulator-5554 (10.8s)
  launch      com.stime2e (834ms)
  verify      process alive 3.2s after launch (release: no bundle fetch to observe)
```

Pulled the installed APK and compared it with the cache entry (`System.IO.Compression`, len/comp/crc):

```
cache entry:   classes.dex len=9921400 comp=3547110 crc=3446907892
               assets/index.android.bundle len=1003716 comp=1003716 crc=4121968217
               resources.arsc len=352740 comp=352740 crc=2836140654
installed:     classes.dex len=9921400 comp=3547110 crc=3446907892
               assets/index.android.bundle len=1268036 comp=1268036 crc=1697700793
               resources.arsc len=352740 comp=352740 crc=2836140654
```

The bundle was replaced and is stored, `resources.arsc` stayed stored, `classes.dex` stayed deflated. `apksigner verify --print-certs` reports the debug signer (V3), `zipalign -c -P 16 4` exits 0.

**Named ports.** `stim ports get web` -> 8900, a `node` listener bound to it, then:

```
> stim ports
web              8900
metro (managed)  8082
> stim ports stop web --dry-run
would stop web (8900): pid 7416 node.exe
would release web (8900)
> stim worktree remove          (with the listener still bound)
stopped web (8900): pid 7104 node.exe
released web (8900)
```

`netstat -ano` showed no 8900 listener afterwards. The `git worktree remove` step of that run then failed with `Permission denied` because my ssh shell's cwd was inside the worktree; rerunning from outside exited 0. Unrelated to this change.

PR #923, opened 2026-09-20
fix: Windows --remote eas follow-ups: doctor host gating, remote collector, autocrlf finding

### LINKED ISSUE #917: Follow-ups from the Windows --remote eas field run
Non-blocking leftovers from the #915 review and field run, none of which block `stim ios --remote eas` on Windows:

- `commands/doctor.ts:209` still calls `detectXcodeMajor()` on non-darwin hosts for `--platform ios`; `checkStorageLayout` still emits DerivedData findings on non-darwin. Both should be gated on the host like the other macOS findings in `runDoctor`.
- The remote log collector spawns `xcrun` on Windows and records a harmless `collector_failed`; it should not attempt an xcrun collector when the device is remote.
- `website/docs/eas-builds.md`: the profile example uses `ios-simulator` while the Windows section uses `development-simulator`, and neither says that `eas-build.ts` requires `ios.simulator: true` for the remote simulator path.
- `supervisor/server-expo.ts`: `expoBinFromPackage(resolvePackageJson(root, 'expo'))` re-reads package.json a second time to decide the win32 spawn; `expoBinPath` could return the source alongside the path.
- Host-side, not stim: `core.autocrlf=true` on a Windows checkout rewrites `ios/.xcode.env` and breaks the EAS build; worth a doctor finding on win32 when the project has that file and the repo has autocrlf on.
- Android `--remote eas` was not exercised from Windows.

janicduplessis:
## Description

Leftovers from the Windows `stim ios --remote eas` field run (#915). Item 6 of #917, Android `--remote eas` from Windows, is not code; its outcome is in a comment below.

- `doctor --platform ios` on Windows or Linux still ran `xcodebuild -version` and reported DerivedData volume findings for a directory that never exists there.
- A remote iOS run spawned the simctl log collector against the session id, which is an EAS session or a daemon host, never a simctl device. It exited at once: `collector_stopped` on macOS, `collector_failed` with ENOENT on Windows.
- A Windows checkout with `core.autocrlf=true` checks `ios/.xcode.env` out with CRLF; the EAS build uploads it that way and the worker's shell fails with `: command not found`. Nothing pointed at the cause.
- `website/docs/eas-builds.md` said `ios-simulator` while the Windows or Linux section said `development-simulator`, and neither mentioned the `ios.simulator: true` requirement `eas-build.ts` enforces.
- `startExpoServer` re-read the expo `package.json` to decide the win32 own-node spawn.

## Solution

`doctorCommand` takes the host the same way it takes `inspectVersions` and passes it to `runDoctor` and `checkStorageLayout`; the Xcode probe runs only on darwin.

A remote iOS launch writes one `collector_skipped` info record saying device logs come from agent-device/EAS instead of spawning `xcrun`, on every host.

A new `cost` finding, `ios/.xcode.env has CRLF line endings`, fires on win32 when the file contains a carriage return. It keys on the bytes rather than on `core.autocrlf` because with autocrlf on the index already holds LF, so a config change alone leaves the CRLF working-tree bytes that eas-cli uploads; the detail names autocrlf as the likely cause and the remedy ends with a checkout of the file (or the `.gitattributes` line).

`resolveExpoBin` returns the source with the path; `expoBinPath` keeps its signature for the other callers.

## Test plan

- `doctor.test.ts`: `doctor --json --platform ios` on a win32 and linux host records no `xcodebuild` call; the CRLF finding appears only for win32 + CR bytes in `ios/.xcode.env`, runs no git command, and stays off for LF, darwin, linux and `--platform android`.
- `doctor-storage.test.ts`: a win32/linux host keeps the shared staging finding and drops the iOS ones.
- `ios-command.test.ts`: a `--remote eas` run calls no `replaceCollector` and writes the `collector_skipped` record.
- `pnpm run format:check`, `lint`, `typecheck`, `knip`, `pnpm test` before each commit.

Fixes #917


janicduplessis:
## Android `--remote eas` from Windows (issue item 6)

Host: `windows-debug.yml` run 35485774836 (windows-latest, stim `main` 8bb4892 dist, eas-cli latest, agent-device 0.20.10, cloudflared via choco). Project: appandflow/trailhead `main` 78c2b74 (SDK 58 preview) cloned to `C:\th`, `npm ci`.

No finished Android `development` build existed, so `eas build --platform android --profile development` was run by hand first: build `9c84e29f`, 03:22:47Z to 03:50:32Z (27m45s). The account showed 97% of included build credits used before this build. It produced a 449 MB APK.

Then, in one SSH session:

- `stim start --remote`: 22s, expo-child on 8082, cloudflared tunnel.
- `stim android --remote eas --eas-profile development`: fingerprint `3f8cc4e5` matched the EAS build, APK downloaded, EAS Android session `01a0bcf1` created and booted (3m17s), then `STIM_INSTALL_FAILED`: `agent-device install` returned `UNAUTHORIZED: Lease is not active`. 5m44s total.

Cause, from agent-device's own diagnostics on both the stim run and a by-hand `agent-device install` of the same APK: the client allocates a lease (`lease_allocate`, daemon default TTL 60s), uploads the 449 MB APK for 1m47s with no heartbeat, then the `install` request finds the lease expired. Not a Windows or stim failure; a large-artifact agent-device client issue (no matching issue in callstack/agent-device). The iOS run in #915 had a small artifact and did not hit it.

`stim stop` ended the session (`stopped remote session 01a0bcf1`), `eas simulator:list` shows it `STOPPED`, no node or cloudflared process remained. EAS cost: one Android build (27m45s, past the included credits) and one Android simulator session of about 10.5 minutes (03:52:38Z to 04:03:07Z).

Two host-side observations, not changed here: on Windows OpenSSH the detached supervisor, Expo and cloudflared processes are killed when the SSH session that started them closes, so `start` and `android` must run in one session; and the trailhead checkout on the runner has `core.autocrlf=true` with CR bytes in `ios/.xcode.env`, which is exactly the case the new doctor finding reports.

